PR #13, opened 2021-06-23
Name wrapper

makoto:
## Features added

- Allow NameWrapper owners to update record
- Replace setAuthorisation (set approval per owner per name) with setApprovalForAll(set approval per owner)

@Arachnid  I ported the validation to disapprove operator to itself but wasn't sure if we need the logic for the resolver as well (just to keep the logic consistent). Do you want me to keep it or get rid of it?


contracts/resolvers/DummyNameWrapper.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity ^0.8.4;
```
Arachnid, line 1: Can we put this in a test directory?

contracts/resolvers/PublicResolver.sol
```diff
@@ -10,47 +10,60 @@ import "./profiles/NameResolver.sol";
...
-    mapping(bytes32=>mapping(address=>mapping(address=>bool))) public authorisations;
+    mapping(address => mapping(address => bool)) private _operatorApprovals;
 
-    event AuthorisationChanged(bytes32 indexed node, address indexed owner, address indexed target, bool isAuthorised);
+    // Logged when an operator is added or removed.
+    event ApprovalForAll(address indexed owner, address indexed operator, bool approved);
 
-    constructor(ENS _ens) {
+    constructor(ENS _ens, INameWrapper wrapperAddress){
+        nameWrapper = wrapperAddress;
```
Arachnid, line 37: Nit: These two lines should be the other way around.

Arachnid (approved):

makoto:
NOTE: This is to solve https://github.com/ensdomains/name-wrapper/issues/26

PR #15, opened 2021-06-28
Controller for Reverse Registrar

jefflau:
Adds Controller pattern to Reverse Registrar and adds a method to allow controllers to set any name to any reverse account node.

This PR will allow the permanent registrar controller (once added as a controller) to setName on behalf of a user
so then we can create a new register method on the permanent registrar controller that can call `setNameForAddr` and setup on the reverse record on behalf of a user. This will allow it to be all set in one tx on a new `register` function

30/6/21 addition:

Added functions as per @Arachnid's request to mirror the original functions so they can all be called with a custom address. Renamed them to `forAddr`.

`forAddr` functions can now be called by the address themselves, controller, or an authorised party on the ENS registry. The reason for adding approvals as it also allows flexibility for contracts to set reverse records on behalf of their users if they have already authorised their contract on the registry. It would also allow them to switch the reverse record if the user created a new subdomain with them in one tx.

Added new tests for all additional functions.

### DIFF
--- a/contracts/registry/ReverseRegistrar.sol
+++ b/contracts/registry/ReverseRegistrar.sol
@@ -83,9 +83,10 @@ contract ReverseRegistrar is Ownable, Controllable {
      * resolver if necessary.
      * Only callable by controllers
      * @param name The name to set for this address.
+     * @param name The name to set for this address.
      * @return The ENS node hash of the reverse record.
      */
-    function setNameWithController(address addr, string memory name)
+    function setNameForAddr(address addr, string memory name)
         public
         onlyController
         returns (bytes32)
@@ -152,10 +153,11 @@ contract ReverseRegistrar is Ownable, Controllable {
         bytes32 label = sha3HexAddress(addr);
         bytes32 node = keccak256(abi.encodePacked(ADDR_REVERSE_NODE, label));
         address currentResolver = ens.resolver(node);
-        address newResolver = (resolver != address(0x0) &&
-            resolver != currentResolver)
-            ? resolver
-            : currentResolver;
+        address newResolver = currentResolver;
+
+        if (resolver != address(0x0) && resolver != currentResolver) {
+            newResolver = resolver;
+        }
 
         ens.setSubnodeRecord(ADDR_REVERSE_NODE, label, owner, newResolver, 0);
 
--- a/contracts/registry/ReverseRegistrar.sol
+++ b/contracts/registry/ReverseRegistrar.sol
@@ -42,7 +42,8 @@ contract ReverseRegistrar is Ownable, Controllable {
         require(
             addr == msg.sender ||
                 controllers[msg.sender] ||
-                ens.isApprovedForAll(addr, msg.sender),
+                ens.isApprovedForAll(addr, msg.sender) ||
+                ownsContract(addr),
             "Caller is not a controller or authorised by address or the address itself"
         );
         _;
@@ -181,13 +182,8 @@ contract ReverseRegistrar is Ownable, Controllable {
         }
     }
 
-    /**
-     * @dev Transfers ownership of the reverse ENS record associated with the
-     *      calling account.
-     * @param owner The address to set as the owner of the reverse record in ENS.
-     * @param resolver The address of the resolver to set; 0 to leave unchanged.
-     * @return The ENS node hash of the reverse record.
-     */
+    /* Internal functions */
+
     function _claimWithResolver(
         address addr,
         address owner,
@@ -205,5 +201,13 @@ contract ReverseRegistrar is Ownable, Controllable {
         emit ReverseClaimed(addr, node);
 
         return node;
+    }
+
+    function ownsContract(address addr) internal view returns (bool) {
+        try Ownable(addr).owner() returns (address owner) {
+            return owner == msg.sender;
+        } catch {
+            return false;
+        }
     }
 }
--- a/contracts/registry/ReverseRegistrar.sol
+++ b/contracts/registry/ReverseRegistrar.sol
@@ -82,18 +82,19 @@ contract ReverseRegistrar is Ownable, Controllable {
      * @param name The name to set for this address.
      * @return The ENS node hash of the reverse record.
      */
-    // function setName(string memory name)
-    //     public
-    //     onlyController
-    //     returns (bytes32)
-    // {
-    //     bytes32 node = claimWithResolver(
-    //         address(this),
-    //         address(defaultResolver)
-    //     );
-    //     defaultResolver.setName(node, name);
-    //     return node;
-    // }
+    function setNameWithController(address addr, string memory name)
+        public
+        onlyController
+        returns (bytes32)
+    {
+        bytes32 node = _claimWithResolver(
+            addr,
+            address(this),
+            address(defaultResolver)
+        );
+        defaultResolver.setName(node, name);
+        return node;
+    }
 
     /**
      * @dev Returns the node hash for a given account's reverse records.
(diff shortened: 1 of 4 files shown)


contracts/registry/ReverseRegistrar.sol
```diff
@@ -116,4 +139,38 @@ contract ReverseRegistrar {
             ret := keccak256(0, 40)
         }
     }
+
+    /**
+     * @dev Transfers ownership of the reverse ENS record associated with the
+     *      calling account.
```
makoto, line 145: Missing `@param addr`. Also can you change the parameter name to something more meaningful?

makoto: maybe `controllerAddress` or something?

test/registry/TestReverseRegistrar.js
```diff
@@ -1,56 +1,82 @@
...
 
     beforeEach(async () => {
         node = namehash.hash(accounts[0].slice(2).toLowerCase() + ".addr.reverse");
+        node2 = namehash.hash(accounts[1].slice(2).toLowerCase() + '.addr.reverse');
+        node3 = namehash.hash(accounts[2].slice(2).toLowerCase() + '.addr.reverse');
         ens = await ENS.new();
         nameWrapper = await NameWrapper.new();
         resolver = await PublicResolver.new(ens.address, nameWrapper.address);
         registrar = await ReverseRegistrar.new(ens.address, resolver.address);
+        await registrar.setController(accounts[0], true)
```
makoto, line 25: Took a while to figure out this is the key logic. You may either want to have this logic in each test where it matters to make it explicit.

test/registry/TestReverseRegistrar.js
```diff
@@ -1,56 +1,82 @@
...
+    })
+
+    it('allows controller to set name records for other accounts', async () => {
+      await registrar.setNameWithController(accounts[1], 'testname', { from: accounts[0], gas: 1000000 })
+      assert.equal(await ens.resolver(node2), resolver.address)
+      assert.equal(await resolver.name(node2), 'testname')
+    })
+
+    it('forbids non-controller from calling setNameWithController', async () => {
+      await exceptions.expectFailure(registrar.setNameWithController(accounts[1], 'testname', { from: accounts[1], gas: 1000000 }));
```
makoto, line 78: If you put the following here, the test becomes more readable.

```
await registrar.setController(accounts[1], false)
```

contracts/registry/ReverseRegistrar.sol
```diff
@@ -77,7 +66,32 @@ contract ReverseRegistrar {
      * @return The ENS node hash of the reverse record.
      */
     function setName(string memory name) public returns (bytes32) {
-        bytes32 node = claimWithResolver(address(this), address(defaultResolver));
+        bytes32 node = claimWithResolver(
```
Arachnid, line 69: Shouldn't this call the underscore-prefixed method?

contracts/registry/ReverseRegistrar.sol
```diff
@@ -77,7 +66,32 @@ contract ReverseRegistrar {
...
+
+    /**
+     * @dev Sets the `name()` record for the reverse ENS record associated with
+     * the account provided. First updates the resolver to the default reverse
+     * resolver if necessary.
+     * Only callable by controllers
+     * @param name The name to set for this address.
+     * @return The ENS node hash of the reverse record.
+     */
+    function setNameWithController(address addr, string memory name)
```
Arachnid, line 85: I think the naming of this needs some workshopping - `claimWithResolver` and `setNameWithController` use 'with' to mean entirely different things.

contracts/registry/ReverseRegistrar.sol
```diff
@@ -102,9 +119,15 @@ contract ReverseRegistrar {
         addr;
```
Arachnid, line 119: We can probably remove these lines; I think Solidity should correctly detect that they're used now.

contracts/registry/ReverseRegistrar.sol
```diff
@@ -102,9 +119,15 @@ contract ReverseRegistrar {
...
         ret; // Stop warning us about unused variables
         assembly {
-            let lookup := 0x3031323334353637383961626364656600000000000000000000000000000000
+            let
```
Arachnid, line 122: Can you make this a global constant (outside the contract scope)?

contracts/registry/ReverseRegistrar.sol
```diff
@@ -116,4 +139,38 @@ contract ReverseRegistrar {
...
+        address addr,
+        address owner,
+        address resolver
+    ) internal returns (bytes32) {
+        bytes32 label = sha3HexAddress(addr);
+        bytes32 node = keccak256(abi.encodePacked(ADDR_REVERSE_NODE, label));
+        address currentOwner = ens.owner(node);
+
+        // Update the resolver if required
+        if (resolver != address(0x0) && resolver != ens.resolver(node)) {
```
Arachnid, line 160: I think most of this logic can be replaced by a call to `setSubnodeRecord`.

contracts/registry/ReverseRegistrar.sol
Arachnid: Also, it'd be useful for the controller to be able to do more than just set the name. Perhaps we can just expose versions of each function with the address parameter, and make them callable only by the address or a controller?

Arachnid (changes requested):

contracts/registry/ReverseRegistrar.sol
```diff
@@ -1,14 +1,19 @@
...
     function setName(bytes32 node, string memory name) public virtual;
 }
 
-contract ReverseRegistrar {
+bytes32 constant lookup = 0x3031323334353637383961626364656600000000000000000000000000000000;
+
+contract ReverseRegistrar is Ownable, Controllable {
     // namehash('addr.reverse')
-    bytes32 public constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
+    bytes32 public constant ADDR_REVERSE_NODE =
```
Arachnid, line 15: Should probably move this outside the contract too.

contracts/registry/ReverseRegistrar.sol
```diff
@@ -18,55 +23,82 @@ contract ReverseRegistrar {
...
+        );
         if (address(oldRegistrar) != address(0x0)) {
             oldRegistrar.claim(msg.sender);
         }
     }
 
+    modifier authorised(address addr) {
+        require(
+            controllers[msg.sender] ||
+                addr == msg.sender ||
```
Arachnid, line 42: For gas-efficiency, we probably want to put this one first.

contracts/registry/ReverseRegistrar.sol
```diff
@@ -116,4 +178,28 @@ contract ReverseRegistrar {
...
     }
+
+    /**
+     * @dev Transfers ownership of the reverse ENS record associated with the
+     *      calling account.
+     * @param owner The address to set as the owner of the reverse record in ENS.
+     * @param resolver The address of the resolver to set; 0 to leave unchanged.
+     * @return The ENS node hash of the reverse record.
+     */
+    function _claimWithResolver(
```
Arachnid, line 189: Can you make this emit an event with at least `addr` and `node?

contracts/registry/ReverseRegistrar.sol
```diff
@@ -1,14 +1,19 @@
 pragma solidity >=0.8.4;
 
 import "./ENS.sol";
+import "../root/Ownable.sol";
```
Arachnid, line 4: This would probably be a good time to switch to using OZ's ownable contract.

Arachnid (changes requested):

contracts/registry/ReverseRegistrar.sol
jefflau: Is this because the lookup on `controllers[msg.sender]` is more costly than a simple equality?

Arachnid (approved):

PR #16, opened 2021-07-05
Add support for providing a resolver when claiming

Arachnid:


contracts/dnsregistrar/DNSRegistrar.sol
```diff
@@ -98,11 +85,39 @@ contract DNSRegistrar {
         claim(name, proof);
     }
 
+    function proveAndClaimWithResolver(bytes memory name, DNSSEC.RRSetWithSignature[] memory input, bytes memory proof, address resolver, address addr) public {
+        proof = oracle.submitRRSets(input, proof);
+        (bytes32 rootNode, bytes32 labelHash, address owner) = _claim(name, proof);
+        require(msg.sender == owner, "Only owner can call proveAndClaimWithResolver");
```
makoto, line 91: We no longer allow anyone else to claim? If this is due to the risk that someone could set a custom resolver to do something malicious, we could also create a function like `proveAndClaimWithDefaultResolver` which takes the default resolver by looking up `resolver.ens.eth`.

makoto:
It's separate from this PR but you look like you committed `package-lock.json`. Can you remove it and use `yarn.lock` instead?

makoto:
Please change [allows the owner of a DNS name to claim it in ENS](https://github.com/ensdomains/ens-contracts/blame/238cfbd223868018080be4e9b2935e6fed61e2f8/test/dnsregistrar/TestDNSRegistrar.js#L37) to `allows anyone to claim on behalf of the owner of a DNS name` and change `await registrar.claim(utils.hexEncodeName('foo.test'), proof)` for `accounts[1]` to claim instead of `accounts[0]` to assert that anyone can call the claim function

contracts/dnsregistrar/DNSRegistrar.sol
```diff
@@ -98,11 +85,39 @@ contract DNSRegistrar {
         claim(name, proof);
     }
 
+    function proveAndClaimWithResolver(bytes memory name, DNSSEC.RRSetWithSignature[] memory input, bytes memory proof, address resolver, address addr) public {
```
makoto, line 88: Need `require(resolver != address(0), "Resolver address is required")`

contracts/dnsregistrar/DNSRegistrar.sol
```diff
@@ -98,11 +85,39 @@ contract DNSRegistrar {
...
+        proof = oracle.submitRRSets(input, proof);
```
makoto, line 89: This and line 84 have input have `proof` variables shadowing. Maybe safer to rename the input one to `_proof` to avoid any unintended effect?

contracts/dnsregistrar/DNSRegistrar.sol
```diff
@@ -98,11 +85,39 @@ contract DNSRegistrar {
...
+        (bytes32 rootNode, bytes32 labelHash, address owner) = _claim(name, proof);
+        require(msg.sender == owner, "Only owner can call proveAndClaimWithResolver");
+        ens.setSubnodeRecord(rootNode, labelHash, owner, resolver, 0);
+        if(addr != address(0)) {
+            bytes32 node = keccak256(abi.encodePacked(rootNode, labelHash));
+            AddrResolver(resolver).setAddr(node, addr);
```
makoto, line 95: Assuming the reverse registrar change happens at the same time, is it worth setting a reverse record as well?

contracts/dnsregistrar/DNSRegistrar.sol
Arachnid: We don't allow it because there's no way to know what the owner wanted unless it's them calling it. If we allow setting the default resolver, anyone could then override a custom resolver by calling this.

contracts/dnsregistrar/DNSRegistrar.sol
Arachnid: We're intentionally replacing the call parameter with the new value.

contracts/dnsregistrar/DNSRegistrar.sol
Arachnid: I'd rather just get this change in than further complicate it.

contracts/dnsregistrar/DNSRegistrar.sol
makoto: I update our keybase conversation. We don't do this as there are edge cases like `anyone who does have a resolver set could then have it replaced with the standard resolver by anyone else` and doing it for only names which haven't set resolver cause another issue of `what if the owner changes? Can you do it? Can only the owner?`

Arachnid:
PTAL!

makoto (approved):

PR #20, opened 2021-07-15
Add deployments using hardhat-deploy

Arachnid:


deployments/mainnet/BaseRegistrarImplementation.json
```diff
@@ -0,0 +1,755 @@
+{
+    "address": "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85",
+    "abi": [
```
makoto, line 3: Do you need to save the entire abi? You should be able to get the abi from the artefacts when you compiled then just attach an address like this.

```
const EnsRegistry = await (await ethers.getContractFactory("ENSRegistry")).attach(registryAddress)
```

contracts/dnssec-oracle/algorithms/EllipticCurve.sol
```diff
@@ -61,7 +61,7 @@ contract EllipticCurve {
     /**
      * @dev Transform affine coordinates into projective coordinates.
      */
-    function toProjectivePoint(uint x0, uint y0) public pure
+    function toProjectivePoint(uint x0, uint y0) internal pure
```
makoto, line 64: Any reason you have to change to internal? I see no harm leaving it public so that people can inspect from Etherscan, etc

Arachnid: As internal, the functions get compiled into the contract that uses the library, instead of having to call to an external library using a DELEGATECALL, so it saves gas.

contracts/dnsregistrar/DNSRegistrar.sol
```diff
@@ -90,11 +90,17 @@ contract DNSRegistrar is IDNSRegistrar {
         proof = oracle.submitRRSets(input, proof);
         (bytes32 rootNode, bytes32 labelHash, address owner) = _claim(name, proof);
         require(msg.sender == owner, "Only owner can call proveAndClaimWithResolver");
-        ens.setSubnodeRecord(rootNode, labelHash, owner, resolver, 0);
         if(addr != address(0)) {
             require(resolver != address(0), "Cannot set addr if resolver is not set");
+            // Set ourselves as the owner so we can set a record on the resolver
```
makoto, line 95: :+1

makoto (approved):

PR #32, opened 2021-07-29
ETH Controller, ReverseRegistrar updates, gas efficiencies, interface updates

jefflau:
- [x]  ReverseRegistrar: Add new events for etherscan
- [x]  ReverseRegistrar: Add controller as trusted contract
- [x] ReverseRegistrar: Refactor ReverseRegistrar to reduced API
- [x]  ETH Controller: Name wrapped automatically on registration
- [x]  ETH Controller: Allow arbitrary records to be set on registration
- [x]  ETH Controller: Registration event split between premium/base cost
- [x] ~~ETH Controller: Remove refund for register/renew and adjust registration to ether given to reduce gas cost. New function does not take duration, but derives it from the msg.value~~
- [x]  ETH Controller: Remove code for refund as it would interfere with how many years a premium name may be registered for
- [x]  ETH Controller: add `rentDuration` function to query duration given an ether amount
- [x]  ETH Controller: Remove ability to set commitmentAge for gas efficiency 
- [x]  ETH Controller/PriceOracles: Make all config variables immutable
- [x]  ETH Controller: Add tests for resolver/records to check failure of incorrect resolver (EoA/non-resolver contract)
- [x]  PublicResolver: Trusted contracts - ETHController/ReverseRegistrar 
- [x]  DNSSEC tests: Fix DNSSEC tests
- [x]  Oracles, Controller: Add interfaces for all contracts for interfaceId
- [x] Bulk Renewal: Fix for new renew function using msg.value
- [x]  BaseRegistrarImplementation: Use IBaseRegistrar and remove BaseRegistrar.sol


contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,136 +3,222 @@ pragma solidity >=0.8.4;
...
-        keccak256("renew(string,uint256)")
-    );
-
-    bytes4 constant private COMMITMENT_WITH_CONFIG_CONTROLLER_ID = bytes4(
-        keccak256("registerWithConfig(string,address,uint256,bytes32,address,address)") ^
-        keccak256("makeCommitmentWithConfig(string,address,bytes32,address,address)")
-    );
+    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+
+    bytes4 private constant INTERFACE_META_ID =
```
Arachnid, line 20: These can be eliminated and replaced with `type(InterfaceName).interfaceId`.

Arachnid (changes requested):

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,136 +3,251 @@ pragma solidity >=0.8.4;
...
-    uint public maxCommitmentAge;
-
-    mapping(bytes32=>uint) public commitments;
-
-    event NameRegistered(string name, bytes32 indexed label, address indexed owner, uint cost, uint expires);
-    event NameRenewed(string name, bytes32 indexed label, uint cost, uint expires);
+    uint256 public minCommitmentAge;
+    uint256 public maxCommitmentAge;
+    ReverseRegistrar reverseRegistrar;
+    INameWrapper nameWrapper;
```
jefflau, line 49: make priceOracle, ReverseRegistrar and Namewrapper public variables

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,136 +3,251 @@ pragma solidity >=0.8.4;
...
 
-    function available(string memory name) public view returns(bool) {
+    function available(string memory name) public view returns (bool) {
         bytes32 label = keccak256(bytes(name));
         return valid(name) && base.available(uint256(label));
     }
 
-    function makeCommitment(string memory name, address owner, bytes32 secret) pure public returns(bytes32) {
-        return makeCommitmentWithConfig(name, owner, secret, address(0), address(0));
+    function makeCommitment(
```
jefflau, line 104: Remove makeCommitmentWithConfig

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,136 +3,251 @@ pragma solidity >=0.8.4;
...
     }
 
     function commit(bytes32 commitment) public {
         require(commitments[commitment] + maxCommitmentAge < block.timestamp);
         commitments[commitment] = block.timestamp;
     }
 
-    function register(string calldata name, address owner, uint duration, bytes32 secret) external payable {
-      registerWithConfig(name, owner, duration, secret, address(0), address(0));
+    function register(
```
jefflau, line 154: Ditch register and rename registerWithConfig => register

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -144,35 +259,94 @@ contract ETHRegistrarController is Ownable {
...
+    function _registerWithResolver(
+        bytes32 label,
+        string calldata name,
+        address owner,
+        uint256 duration,
+        address resolver,
+        address addr,
+        uint96 fuses
+    ) internal returns (uint256 expires) {
+        Resolver resolverContract = Resolver(resolver);
```
jefflau, line 333: move this to an if statement for having a resolver

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -144,35 +259,94 @@ contract ETHRegistrarController is Ownable {
...
+        returns (bool)
+    {
+        return
+            interfaceID == INTERFACE_META_ID ||
+            interfaceID == COMMITMENT_CONTROLLER_ID ||
+            interfaceID == COMMITMENT_WITH_CONFIG_CONTROLLER_ID;
     }
 
-    function _consumeCommitment(string memory name, uint duration, bytes32 commitment) internal returns (uint256) {
+    function onERC1155Received(
```
jefflau, line 285: Remove

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -144,35 +259,94 @@ contract ETHRegistrarController is Ownable {
...
-        return interfaceID == INTERFACE_META_ID ||
-               interfaceID == COMMITMENT_CONTROLLER_ID ||
-               interfaceID == COMMITMENT_WITH_CONFIG_CONTROLLER_ID;
+    function supportsInterface(bytes4 interfaceID)
+        external
+        pure
+        returns (bool)
+    {
+        return
+            interfaceID == INTERFACE_META_ID ||
```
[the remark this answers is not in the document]
jefflau: no need for separate commit_with_config_controller_id

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,136 +3,251 @@ pragma solidity >=0.8.4;
...
-
-    bytes4 constant private COMMITMENT_WITH_CONFIG_CONTROLLER_ID = bytes4(
-        keccak256("registerWithConfig(string,address,uint256,bytes32,address,address)") ^
-        keccak256("makeCommitmentWithConfig(string,address,bytes32,address,address)")
-    );
+    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+
+    bytes4 private constant INTERFACE_META_ID =
+        bytes4(keccak256("supportsInterface(bytes4)"));
+    bytes4 private constant COMMITMENT_CONTROLLER_ID =
```
jefflau, line 24: type(commitmentController).interfaceId

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -144,35 +259,94 @@ contract ETHRegistrarController is Ownable {
...
 
-        uint cost = rentPrice(name, duration);
+        uint256 cost = rentPrice(name, duration);
         require(duration >= MIN_REGISTRATION_DURATION);
         require(msg.value >= cost);
 
         return cost;
     }
+
+    function _registerWithResolver(
```
jefflau, line 324: remove

contracts/ethregistrar/BulkRenewal.sol
```diff
@@ -6,43 +6,62 @@ import "./ETHRegistrarController.sol";
...
-    bytes32 constant private ETH_NAMEHASH = 0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae;
-    bytes4 constant private REGISTRAR_CONTROLLER_ID = 0x018fac06;
-    bytes4 constant private INTERFACE_META_ID = bytes4(keccak256("supportsInterface(bytes4)"));
-    bytes4 constant public BULK_RENEWAL_ID = bytes4(
-        keccak256("rentPrice(string[],uint)") ^
-        keccak256("renewAll(string[],uint")
-    );
+    bytes32 private constant ETH_NAMEHASH =
+        0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae;
+    bytes4 private constant REGISTRAR_CONTROLLER_ID = 0x018fac06;
```
Arachnid, line 11: We can replace these with interfaces and `type(interface).interfaceId` now.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
 import "../resolvers/Resolver.sol";
+import "../registry/ReverseRegistrar.sol";
+import "../dnssec-oracle/BytesUtils.sol";
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@ensdomains/name-wrapper/interfaces/INameWrapper.sol";
+
+import "hardhat/console.sol";
+
+interface commitmentController {
```
Arachnid, line 15:
```suggestion
interface CommitmentController {
```

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
-        keccak256("renew(string,uint256)")
+    using BytesUtils for bytes;
+
+    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+
+    bytes4 private constant INTERFACE_META_ID =
+        bytes4(keccak256("supportsInterface(bytes4)"));
+    bytes4 private constant COMMITMENT_CONTROLLER_ID =
+        type(commitmentController).interfaceId;
+    bytes4 private constant COMMITMENT_WITH_CONFIG_CONTROLLER_ID =
```
Arachnid, line 51: I think this is the version we want an interface for; we're no longer supporting the version without config.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
-        keccak256("register(string,address,uint256,bytes32)") ^
-        keccak256("renew(string,uint256)")
+    using BytesUtils for bytes;
+
+    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+
+    bytes4 private constant INTERFACE_META_ID =
+        bytes4(keccak256("supportsInterface(bytes4)"));
+    bytes4 private constant COMMITMENT_CONTROLLER_ID =
+        type(commitmentController).interfaceId;
```
Arachnid, line 50: You can put this inline in `supportsInterface`; no need for a separate constant.

contracts/ethregistrar/ETHRegistrarController.sol
Arachnid: This should be in its own file, too.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
+                keccak256(
+                    "makeCommitment(string,address,bytes32,address,address,bool,int96)"
+                )
+        );
+
+    BaseRegistrarImplementation immutable base;
+    PriceOracle public prices;
+    uint256 public immutable minCommitmentAge;
+    uint256 public immutable maxCommitmentAge;
+    ReverseRegistrar public immutable reverseRegistrar;
```
Arachnid, line 65: Can you extract interfaces for `BaseRegistrarImplementation` and `ReverseRegistrar` too?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
         base = _base;
         prices = _prices;
         minCommitmentAge = _minCommitmentAge;
         maxCommitmentAge = _maxCommitmentAge;
+        reverseRegistrar = _reverseRegistrar;
+        nameWrapper = _nameWrapper;
     }
 
-    function rentPrice(string memory name, uint duration) view public returns(uint) {
+    function rentPrice(string memory name, uint256 duration)
```
Arachnid, line 104: Should we add a pass-through for the inverse too, now that we have it?

Arachnid: Or maybe make this internal, and people can just fetch the PriceOracle through the public variable.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
+        Cost memory cost = rentPrice(name, duration);
+        require(msg.value >= cost.base);
 
         bytes32 label = keccak256(bytes(name));
-        uint expires = base.renew(uint256(label), duration);
+        uint256 expires = base.renew(uint256(label), duration);
 
-        if(msg.value > cost) {
-            payable(msg.sender).transfer(msg.value - cost);
+        if (msg.value > cost.base) {
```
Arachnid, line 217: Shouldn't we do the same as for registration, and just consume whatever was sent?

contracts/ethregistrar/ETHRegistrarController.sol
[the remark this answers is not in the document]
Arachnid: :point_up:

contracts/ethregistrar/ETHRegistrarController.sol
Arachnid: :point_up:

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
 
         return cost;
     }
+
+    function _setRecords(
+        address resolver,
+        bytes32 label,
+        bytes[] calldata data
+    ) internal {
+        // use hardcoded .eth namehash
```
Arachnid, line 288: We should do this and save a call to `base`.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
+        bytes32 nodehash = keccak256(abi.encodePacked(base.baseNode(), label));
+        for (uint256 i = 0; i < data.length; i++) {
+            // check first few bytes are namehash
+            bytes32 txNamehash = data[i].readBytes32(4);
```
Arachnid, line 292: In solidity 0.8, the `readX` methods are obsolete, because we can do this:

```suggestion
            bytes32 txNamehash = bytes32(data[i][4:36])
```

contracts/ethregistrar/LinearPremiumPriceOracle.sol
```diff
@@ -6,35 +6,45 @@ import "./StablePriceOracle.sol";
 contract LinearPremiumPriceOracle is StablePriceOracle {
     using SafeMath for *;
 
-    uint GRACE_PERIOD = 90 days;
+    uint256 GRACE_PERIOD = 90 days;
```
Arachnid, line 9: This should be a constant.

contracts/ethregistrar/LinearPremiumPriceOracle.sol
```diff
@@ -6,35 +6,45 @@ import "./StablePriceOracle.sol";
...
 
-    uint public initialPremium;
-    uint public premiumDecreaseRate;
+    uint256 public immutable initialPremium;
+    uint256 public immutable premiumDecreaseRate;
 
-    bytes4 constant private TIME_UNTIL_PREMIUM_ID = bytes4(keccak256("timeUntilPremium(uint,uint"));
+    bytes4 private constant TIME_UNTIL_PREMIUM_ID =
+        bytes4(keccak256("timeUntilPremium(uint,uint"));
```
Arachnid, line 15: This should be an interface.

contracts/ethregistrar/LinearPremiumPriceOracle.sol
```diff
@@ -6,35 +6,45 @@ import "./StablePriceOracle.sol";
...
+    ) internal view override returns (uint256) {
         expires = expires.add(GRACE_PERIOD);
-        if(expires > block.timestamp) {
+        if (expires > block.timestamp) {
             // No premium for renewals
             return 0;
         }
 
         // Calculate the discount off the maximum premium
-        uint discount = premiumDecreaseRate.mul(block.timestamp.sub(expires));
```
Arachnid, line 31: In 0.8, we can replace all safemath stuff with regular arithmetic, because safe math is the default now.

contracts/ethregistrar/PriceOracle.sol
```diff
@@ -1,12 +1,27 @@
 pragma solidity >=0.8.4;
 
+struct Cost {
```
Arachnid, line 3: Is there a reason to use a struct here instead of multiple return values?

contracts/ethregistrar/StablePriceOracle.sol
```diff
@@ -6,44 +6,72 @@ import "./StringUtils.sol";
...
     AggregatorInterface public usdOracle;
 
     event OracleChanged(address oracle);
 
-    event RentPriceChanged(uint[] prices);
+    event RentPriceChanged(uint256[] prices);
 
-    bytes4 constant private INTERFACE_META_ID = bytes4(keccak256("supportsInterface(bytes4)"));
-    bytes4 constant private ORACLE_ID = bytes4(keccak256("price(string,uint256,uint256)") ^ keccak256("premium(string,uint256,uint256)"));
+    bytes4 private constant INTERFACE_META_ID =
```
Arachnid, line 27: These should be interfaces.

contracts/ethregistrar/StablePriceOracle.sol
```diff
@@ -6,44 +6,72 @@ import "./StringUtils.sol";
...
 
-
 // StablePriceOracle sets a price in USD, based on an oracle.
 contract StablePriceOracle is Ownable, PriceOracle {
     using SafeMath for *;
     using StringUtils for *;
 
     // Rent in base price units by length. Element 0 is for 1-length names, and so on.
-    uint[] public rentPrices;
+    uint256[] public rentPrices;
```
Arachnid, line 18: I think we probably want to replace this with a hardcoded if-else statement for lengths 3, 4, and 5+. That lets us make the values immutable and save another chunk of gas.

contracts/resolvers/DefaultReverseResolver.sol
```diff
@@ -9,32 +9,36 @@ import "../registry/ReverseRegistrar.sol";
...
+    constructor(ENS ensAddr, address reverseRegistrar) {
         ens = ensAddr;
+        owner = reverseRegistrar;
 
         // Assign ownership of the reverse record to our deployer
-        ReverseRegistrar registrar = ReverseRegistrar(ens.owner(ADDR_REVERSE_NODE));
-        if (address(registrar) != address(0x0)) {
-            registrar.claim(msg.sender);
-        }
+        // ReverseRegistrar registrar = ReverseRegistrar(reverseRegistrar);
```
Arachnid, line 38: ?

contracts/resolvers/DefaultReverseResolver.sol
```diff
@@ -44,5 +48,6 @@ contract DefaultReverseResolver {
      */
     function setName(bytes32 node, string memory _name) public onlyOwner(node) {
```
Arachnid, line 49: Weren't we going to change to using the public resolver here?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -3,176 +3,312 @@ pragma solidity >=0.8.4;
...
+    function _setReverseRecord(
+        string calldata name,
+        address resolver,
+        address owner
+    ) internal {
+        reverseRegistrar.setNameForAddr(
+            msg.sender,
+            msg.sender,
+            resolver,
+            string(abi.encodePacked(name, ".eth"))
```
Arachnid, line 311: Spitballing here, but what if we modified the resolver interface to support DNS-encoded names, and made the existing method decode it to human-readable? That would make stuff that needs to read and hash reverse record names onchain much more performant.

Arachnid (changes requested):

contracts/resolvers/DefaultReverseResolver.sol
jefflau: I can't remember the exact reason we still have the reverse resolver. I think it was for setting the reverse for the reverse registrar itself?

We pass through the public resolver in setNameForAddr so we are using the public resolver

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
 import "./BaseRegistrarImplementation.sol";
 import "./StringUtils.sol";
-import "@openzeppelin/contracts/access/Ownable.sol";
 import "../resolvers/Resolver.sol";
+import "../registry/ReverseRegistrar.sol";
+import "../dnssec-oracle/BytesUtils.sol";
+import "./IETHRegistrarController.sol";
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@ensdomains/name-wrapper/interfaces/INameWrapper.sol";
```
makoto, line 11: Why not use https://github.com/ensdomains/ens-contracts/blob/master/contracts/wrapper/INameWrapper.sol instead?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
+
+import "hardhat/console.sol";
```
makoto, line 13: Remove this

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
+        uint256 expires = nameWrapper.registerAndWrapETH2LD(
+            name,
+            owner,
+            duration,
+            resolver,
+            fuses
+        );
+
+        if (data.length > 0) {
+            require(
```
makoto, line 167: What about moving this logic inside `makeCommitment` function so that it throws an error on step1 if the integrating dapp/wallet does not pass resolver info. Currently, it only throws an error at step 2

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
-    function setCommitmentAges(uint _minCommitmentAge, uint _maxCommitmentAge) public onlyOwner {
-        minCommitmentAge = _minCommitmentAge;
-        maxCommitmentAge = _maxCommitmentAge;
+    // function setCommitmentAges(
+    //     uint256 _minCommitmentAge,
+    //     uint256 _maxCommitmentAge
+    // ) public onlyOwner {
+    //     minCommitmentAge = _minCommitmentAge;
+    //     maxCommitmentAge = _maxCommitmentAge;
+    // }
```
makoto, line 217: Remove comments if not needed

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
-    function supportsInterface(bytes4 interfaceID) external pure returns (bool) {
-        return interfaceID == INTERFACE_META_ID ||
-               interfaceID == COMMITMENT_CONTROLLER_ID ||
-               interfaceID == COMMITMENT_WITH_CONFIG_CONTROLLER_ID;
+    function onERC1155Received(
+        address,
+        address,
+        uint256,
+        uint256,
+        bytes calldata
```
makoto, line 238: `calledata` is not used anywhere

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
+                "ETHRegistrarController: resolver is required when data is supplied"
+            );
+            _setRecords(resolver, label, data);
         }
 
-        emit NameRegistered(name, label, owner, cost, expires);
-
-        // Refund any extra payment
-        if(msg.value > cost) {
-            payable(msg.sender).transfer(msg.value - cost);
```
makoto, line 124: why do we no longer do this at registration? We set 10% higher to accommodate for price oracle change

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,299 @@
...
         // If the commitment is too old, or the name is registered, stop
         require(commitments[commitment] + maxCommitmentAge > block.timestamp);
         require(available(name));
 
-        delete(commitments[commitment]);
+        delete (commitments[commitment]);
 
-        uint cost = rentPrice(name, duration);
         require(duration >= MIN_REGISTRATION_DURATION);
-        require(msg.value >= cost);
```
makoto, line 174: why do we get rid of this check?

package.json
```diff
@@ -42,6 +43,7 @@
   },
   "dependencies": {
     "@ensdomains/buffer": "^0.0.13",
+    "@ensdomains/name-wrapper": "^1.0.0",
```
makoto, line 46: as mentioned earlier, it should use the ones migrated over

makoto:
@jefflau can you write down the high-level overview of features you added and changes you made so that I can focus on checking if it has enough test coverage.

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: We get rid of this check to change the logic of the contract to just register how much they sent in ETH. You can see we check a new function from the oracle called `duration` which allows us to figure out how long they want to register for

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: See below:

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: Didn't seem to work for me. I got:

```
TypeError: Explicit type conversion not allowed from "bytes calldata slice" to "bytes32".
bytes32 txNamehash = bytes32(data[i][4:36]);
```

jefflau:
> @jefflau can you write down the high-level overview of features you added and changes you made so that I can focus on checking if it has enough test coverage.

Done.

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: I'm not sure if github is showing the right code snippet. But if it's L238, seems to be used

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: Added

test/registry/TestReverseRegistrar.js
```diff
@@ -127,191 +167,216 @@ describe('ReverseRegistrar Tests', () => {
         assertReverseClaimedEventEmitted(tx, accounts[0], node)
       })
 
-      it('does not overwrite resolver if not specified', async () => {
-        await registrar.claimWithResolver(accounts[1], accounts[2], {
-          from: accounts[0],
-        })
-        await registrar.claim(accounts[3], { from: accounts[0] })
+      // it('does not overwrite resolver if not specified', async () => {
```
makoto, line 170: Please either fix the test or delete if not needed any more

test/registry/TestReverseRegistrar.js
```diff
@@ -127,191 +167,216 @@ describe('ReverseRegistrar Tests', () => {
...
-          accounts[0],
-          resolver.address,
-          {
-            from: accounts[0],
-          }
-        )
-        assert.equal(await ens.owner(node), accounts[0])
-        assert.equal(await ens.resolver(node), resolver.address)
+    // describe('claimWithResolverForAddr', () => {
+    //   it('allows an account with the same address to specify resolver', async () => {
```
makoto, line 182: Please either fix the test or delete if not needed any more

test/registry/TestReverseRegistrar.js
```diff
@@ -127,191 +167,216 @@ describe('ReverseRegistrar Tests', () => {
...
+        await registrar.setController(accounts[0], true)
+        await registrar.setNameForAddr(
+          accounts[1],
           accounts[0],
           resolver.address,
+          'testname',
           {
-            from: accounts[1],
+            from: accounts[0],
+            gas: 1000000,
```
makoto, line 317: why do you need to set gas at individual test? Can you not adjust at hardhat config or somewhere global config?

contracts/resolvers/DefaultReverseResolver.sol
```diff
@@ -9,32 +9,30 @@ import "../registry/ReverseRegistrar.sol";
...
      */
-    constructor(ENS ensAddr) {
+    constructor(ENS ensAddr, address reverseRegistrar) {
         ens = ensAddr;
-
-        // Assign ownership of the reverse record to our deployer
-        ReverseRegistrar registrar = ReverseRegistrar(ens.owner(ADDR_REVERSE_NODE));
-        if (address(registrar) != address(0x0)) {
-            registrar.claim(msg.sender);
-        }
```
makoto, line 37: Which test does it cover this change?

contracts/resolvers/PublicResolver.sol
```diff
@@ -51,9 +73,15 @@ contract PublicResolver is ABIResolver, AddrResolver, ContentHashResolver, DNSRe
         emit ApprovalForAll(msg.sender, operator, approved);
     }
 
-    function isAuthorised(bytes32 node) internal override view returns(bool) {
+    function isAuthorised(bytes32 node) internal view override returns (bool) {
+        if (
```
makoto, line 77: Can you point me to which tests cover these logics? I couldn't find one

contracts/ethregistrar/StablePriceOracle.sol
```diff
@@ -4,95 +4,138 @@ import "./PriceOracle.sol";
...
-    uint[] public rentPrices;
+    // Rent in base price units by length
+    uint256 public immutable price1Letter;
+    uint256 public immutable price2Letter;
+    uint256 public immutable price3Letter;
+    uint256 public immutable price4Letter;
+    uint256 public immutable price5Letter;
 
     // Oracle address
     AggregatorInterface public usdOracle;
```
makoto, line 26: Can this also be immutable?

contracts/ethregistrar/BulkRenewal.sol
```diff
@@ -3,46 +3,67 @@ pragma experimental ABIEncoderV2;
...
+import "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+
+contract BulkRenewal is IBulkRenewal {
+    bytes32 private constant ETH_NAMEHASH =
+        0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae;
 
     ENS public ens;
 
-    constructor(ENS _ens) public {
+    constructor(ENS _ens) {
```
makoto, line 18: ` ENS public ens;` can be immutable

contracts/registry/TestRegistrar.sol
```diff
@@ -7,11 +7,11 @@ import "./ENS.sol";
  * expires registrations a fixed period after they're initially claimed.
  */
 contract TestRegistrar {
-    uint constant registrationPeriod = 4 weeks;
+    uint256 constant registrationPeriod = 4 weeks;
 
     ENS public ens;
```
makoto, line 12: this can be immutable

contracts/ethregistrar/BaseRegistrar.sol
```diff
@@ -1,51 +1,32 @@
...
+    );
+    event NameRegistered(
+        uint256 indexed id,
+        address indexed owner,
+        uint256 expires
+    );
+    event NameRenewed(uint256 indexed id, uint256 expires);
 
     // The ENS registry
     ENS public ens;
```
makoto, line 26: Maybe this can be immutable?

contracts/ethregistrar/ETHRegistrarController.sol
makoto: What am I seeing below?

makoto:
```
ReverseRegistrar: Add new events for etherscan
```
Which events are for etherscan?

contracts/ethregistrar/BaseRegistrar.sol
Arachnid: We can't redeploy this, so we shouldn't make meaningful changes to it.

contracts/ethregistrar/StablePriceOracle.sol
jefflau: Yes! I must have left that out whilst testing gas efficiency. Good catch.

contracts/registry/TestRegistrar.sol
jefflau: i'm not sure we care about the test registrar, but i'll add it anyway

contracts/resolvers/DefaultReverseResolver.sol
```diff
@@ -9,32 +9,30 @@ import "../registry/ReverseRegistrar.sol";
  */
 contract DefaultReverseResolver {
     // namehash('addr.reverse')
-    bytes32 constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
+    bytes32 constant ADDR_REVERSE_NODE =
+        0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
 
     ENS public ens;
```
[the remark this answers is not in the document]
jefflau: I've left this resolver in, but I don't think we're using it anymore. We're now using the public resolver as the default

contracts/resolvers/DefaultReverseResolver.sol
jefflau: Default reverse resolver isn't used anymore. Going to delete

contracts/ethregistrar/BulkRenewal.sol
jefflau: > ```
> ReverseRegistrar: Add new events for etherscan
> ```
> 
> Which events are for etherscan?



> ```
> ReverseRegistrar: Add new events for etherscan
> ```
> 
> Which events are for etherscan?

Actually this was added in a previous PR, it's called `ReverseClaimed`

jefflau: Added!

contracts/registry/TestRegistrar.sol
jefflau: Edit, I just realised this is a constant, which is just as good as immutable since it is never set, but hardcoded

jefflau:
Addressed your comments @makoto

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,275 @@
...
+            (bool success, ) = address(resolver).call(data[i]);
+            require(success, "ETHRegistrarController: Failed to set Record");
+        }
+    }
 
-        return cost;
+    function _setReverseRecord(
+        string calldata name,
+        address resolver,
+        address owner
```
makoto, line 266: `address owner` variable is unused

test/ethregistrar/TestEthRegistrarController.js
```diff
@@ -1,185 +1,727 @@
...
     it('should allow the registrar owner to withdraw funds', async () => {
-        await controller.withdraw({from: ownerAccount});
-        assert.equal(await web3.eth.getBalance(controller.address), 0);
-    });
-});
+      await controller.withdraw({ from: ownerAccount })
+      assert.equal(await web3.eth.getBalance(controller.address), 0)
+    })
+
+    it('should set the reverse record of the account', async () => {
```
makoto, line 545: Can you add the test when `reverseRecord` is set to `false`?

test/ethregistrar/TestEthRegistrarController.js
```diff
@@ -1,185 +1,727 @@
...
+      assert.equal(
+        await baseRegistrar.ownerOf(sha3('newconfigname')),
+        nameWrapper.address
+      )
+      assert.equal(await resolver['addr(bytes32)'](nodehash), registrantAccount)
+      assert.equal(await resolver['text'](nodehash, 'url'), 'ethereum.com')
+      assert.equal(await nameWrapper.ownerOf(nodehash), registrantAccount)
+    })
+
+    it('should not permit new registrations with records updating a different name', async () => {
```
makoto, line 277: You can add another test case making sure `ETHRegistrarController: Failed to set Record` (maybe pass empty  or wrong resolver address)

test/ethregistrar/TestEthRegistrarController.js
```diff
@@ -1,185 +1,727 @@
...
+      )
+    })
 
     it('should require sufficient value for a renewal', async () => {
-        await exceptions.expectFailure(controller.renew("name", 86400));
-    });
+      await exceptions.expectFailure(controller.renew('name'))
+    })
 
     it('should allow the registrar owner to withdraw funds', async () => {
```
makoto, line 540: probably more correct to say "should allow anyone to withdraw funds and transfer to the registrar owner"

contracts/ethregistrar/BaseRegistrar.sol
makoto: Do you mean we can't redeploy `BaseRegistrar.sol`? If so, why do we make any changes to the file at all?

makoto:
Added more comments. Also there are several comments seems not addressed

contracts/ethregistrar/BulkRenewal.sol
```diff
@@ -3,46 +3,67 @@ pragma experimental ABIEncoderV2;
...
+        external
+        view
+        override
+        returns (uint256 total)
+    {
         ETHRegistrarController controller = getController();
-        for(uint i = 0; i < names.length; i++) {
-            total += controller.rentPrice(names[i], duration);
+        for (uint256 i = 0; i < names.length; i++) {
+            (uint256 basePrice, ) = controller.rentPrice(names[i], duration);
```
Arachnid, line 41: Won't this return the wrong value? It seems like it should sum the premiums as well, either as part of the total or as a separate field.

contracts/ethregistrar/BulkRenewal.sol
```diff
@@ -3,46 +3,67 @@ pragma experimental ABIEncoderV2;
...
+        external
+        payable
+        override
+    {
         ETHRegistrarController controller = getController();
-        for(uint i = 0; i < names.length; i++) {
-            uint cost = controller.rentPrice(names[i], duration);
-            controller.renew{value:cost}(names[i], duration);
+        for (uint256 i = 0; i < names.length; i++) {
+            (uint256 basePrice, ) = controller.rentPrice(names[i], duration);
```
Arachnid, line 53: As above - won't this send the wrong amount for any name with a premium on it?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,275 @@
...
-        keccak256("register(string,address,uint256,bytes32)") ^
-        keccak256("renew(string,uint256)")
+    using BytesUtils for bytes;
+
+    uint256 public constant MIN_REGISTRATION_DURATION = 28 days;
+    bytes32 private constant ETH_NODE =
+        0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae;
+
+    BaseRegistrarImplementation immutable base;
+    PriceOracle public prices;
```
Arachnid, line 26: Where did we come down on making this immutable? Deploying a new controller for each oracle change would be a pain, but it would cut a bit of gas off every registration tx.

contracts/ethregistrar/ETHRegistrarController.sol
Arachnid: PTAL!

contracts/ethregistrar/ETHRegistrarController.sol
Arachnid: :point_up:

contracts/ethregistrar/StablePriceOracle.sol
```diff
@@ -4,86 +4,138 @@ import "./PriceOracle.sol";
...
+
+        if (len == 1) {
+            basePrice = price1Letter * duration;
+        } else if (len == 2) {
+            basePrice = price2Letter * duration;
+        } else if (len == 3) {
+            basePrice = price3Letter * duration;
+        } else if (len == 4) {
+            basePrice = price4Letter * duration;
+        } else if (len == 5) {
```
Arachnid, line 60: Um, what if the length is >5?

A unit test should definitely have caught this!

contracts/registry/ReverseRegistrar.sol
```diff
@@ -49,14 +48,19 @@ contract ReverseRegistrar is Ownable, Controllable {
...
     /**
      * @dev Transfers ownership of the reverse ENS record associated with the
      *      calling account.
      * @param owner The address to set as the owner of the reverse record in ENS.
      * @return The ENS node hash of the reverse record.
      */
-    function claim(address owner) public returns (bytes32) {
-        return _claimWithResolver(msg.sender, owner, address(0x0));
+    function claim(address owner) public override returns (bytes32) {
+        require(address(defaultResolver) != address(0));
```
Arachnid, line 62: We can remove this check, and just not allow it to be set to 0 in the first place.

Arachnid (commented):
There's also some comments from @makoto that need addressing.

contracts/ethregistrar/StablePriceOracle.sol
jefflau:
```
if (len > 5) {
    len = 5;
}
```

We normalise the value at the top (technically you did this)

contracts/ethregistrar/BulkRenewal.sol
jefflau: Names that aren't expired don't have premiums?

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: I think we can make it immutable. I don't think we'll end up changing the oracle often if we can sort out this exponential oracle out before all these contracts are ready

contracts/ethregistrar/BulkRenewal.sol
jefflau: As above, names that can be renewed shouldn't have premiums

test/registry/TestReverseRegistrar.js
jefflau: deleted these. I think they were set to 0 causing tests to fail, so i set it higher, but i think the default should work

contracts/resolvers/PublicResolver.sol
jefflau: Added one

jefflau:
```
it('trusted contract can bypass authorisation', async () => {
        await resolver.methods['setAddr(bytes32,address)'](node, accounts[9], {
          from: accounts[9],
        })
        assert.equal(await resolver.addr(node), accounts[9])
      })
```

contracts/ethregistrar/StablePriceOracle.sol
Arachnid: :facepalm:

contracts/ethregistrar/BulkRenewal.sol
Arachnid: Excellent point. Perhaps we could sum it, or require() that it's zero, though? We don't want a situation where we use the premium for something and accidentally forget to charge it.

jefflau: Happy to sum it since it will just be 0. Leaning towards summing than require it is 0, since our tests should show a premium being incorrectly calculated if we change the premium oracle and accidentally let a premium slip into renews

contracts/registry/ReverseRegistrar.sol
jefflau: Do you mean not allow it to be set to 0 in the UI or you mean further down inside `claimForAddr`?

test/ethregistrar/TestEthRegistrarController.js
jefflau: Done. Added one for an incorrect resolver and an EoA. `0x0...` resolver gets caught earlier in the makeCommitment, which I also added a test for

test/ethregistrar/TestStablePriceOracle.js
```diff
@@ -1,36 +1,81 @@
...
+        (await priceOracle.duration('fubar', 1, 1000000n))[0].toNumber()
+      ).to.equal(100000)
+
+      expect(
+        (await priceOracle.duration('foobie', 1, 1000000n))[0].toNumber()
+      ).to.equal(100000)
     })
-});
+
+    // it('should work with larger values', async () => {
```
makoto, line 72: Why is this test commented out? Can we either uncomment or delete if not needed?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -142,24 +147,27 @@ contract ETHRegistrarController is Ownable, IETHRegistrarController {
...
     ) public payable override {
         bytes32 label = keccak256(bytes(name));
-        (uint256 duration, uint256 premiumCost) = prices.duration(
-            name,
-            base.nameExpires(uint256(label)),
-            msg.value
+        uint256[2] memory cost = rentPrice(name, duration);
+        require(
+            msg.value >= (cost[0] + cost[1]),
+            "ETHRegistrarController: Not enough ether provided"
```
makoto, line 158: Can you add test for this logic?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -182,14 +190,11 @@ contract ETHRegistrarController is Ownable, IETHRegistrarController {
...
-            label,
-            owner,
-            msg.value - premiumCost,
-            premiumCost,
-            expires
-        );
+        emit NameRegistered(name, label, owner, cost[0], cost[1], expires);
+
+        if (msg.value > (cost[0] + cost[1])) {
+            payable(msg.sender).transfer(msg.value - (cost[0] + cost[1]));
```
makoto, line 193: Is there a test for this logic? (I couldn't find it).

contracts/ethregistrar/BulkRenewal.sol
```diff
@@ -3,46 +3,73 @@ pragma experimental ABIEncoderV2;
...
         ETHRegistrarController controller = getController();
-        for(uint i = 0; i < names.length; i++) {
-            uint cost = controller.rentPrice(names[i], duration);
-            controller.renew{value:cost}(names[i], duration);
+        for (uint256 i = 0; i < names.length; i++) {
+            (uint256 basePrice, uint256 premium) = controller.rentPrice(
+                names[i],
+                duration
+            );
+            controller.renew{value: basePrice + premium}(names[i]);
```
makoto, line 60: People should be paying premium only at registration, not for renewal.  Maybe add `require(premium == 0)` instead?

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: I believe the logic where it tests the balance before and after in the contract tests this. Did you have a suggestion on how to test it?

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: Will do

test/ethregistrar/TestStablePriceOracle.js
jefflau: Added this back in with a different test. Was removed since we didn't have that function

contracts/ethregistrar/BaseRegistrar.sol
```diff
@@ -1,51 +1,32 @@
 pragma solidity ^0.8.4;
```
Arachnid, line 1: Do we actually need this file anymore? Can't we move the constant and events into `IBaseRegistrar`, and the mapping into `BaseRegistrarImplementation`?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,178 +1,278 @@
...
-        require(commitments[commitment] + maxCommitmentAge > block.timestamp);
-        require(available(name));
+        require(
+            commitments[commitment] + maxCommitmentAge > block.timestamp,
+            "ETHRegistrarController: Commitment has expired"
+        );
+        require(available(name), "ETHRegistrarController: Name is unavailable");
 
-        delete(commitments[commitment]);
+        delete (commitments[commitment]);
```
Arachnid, line 240: Why the space?

contracts/ethregistrar/LinearPremiumPriceOracle.sol
```diff
@@ -1,40 +1,46 @@
 pragma solidity >=0.8.4;
```
Arachnid, line 1: Assuming it still compiles, I think we can just revert all the changes to this file, since we're not going to deploy the modified contract.

test/registry/TestReverseRegistrar.js
```diff
@@ -126,192 +163,91 @@ describe('ReverseRegistrar Tests', () => {
...
+            'testname',
             {
-              from: accounts[1],
+              from: accounts[0],
             }
           )
         )
       })
 
-      it('allows an owner() of a contract to claimWithResolverForAddr on behalf of the contract', async () => {
```
Arachnid, line 234: Why were these tests deleted? Do we have any others that cover these cases?

Arachnid (changes requested):

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: My prettier auto added that on a fresh install of VSCode + solidity plugin. I'm not sure how to get rid of it

contracts/ethregistrar/BaseRegistrar.sol
jefflau: I guess so, but since BaseRegistrarImplementation is already deployed and likely never to be redeployed - is it worth making these changes? If you think so I can change it

jefflau:
Resolved all comments. Ready for merging imo.

Arachnid (approved):

CryptoGodfatherVA2 (commented):
var5

Issue #41, opened 2021-09-27
Write and deploy universal resolver


Arachnid:
Here's an outline:
```
interface ENS {
    function resolver(bytes32 node) external view returns(address);
}

contract UniversalResolver {
    ENS ens;
    
    constructor(ENS _ens) {
        ens = _ens;
    }
    
    function multicall(bytes[] calldata inputs) external returns (bytes[] memory ret) {
        ret = new bytes[](inputs.length);
        for(uint256 i = 0; i < inputs.length; i++) {
            (, bytes memory result) = _resolve(inputs[i]);
            ret[i] = result;
        }
        return ret;
    }
    
    fallback (bytes calldata input) external returns (bytes memory _output) {
        (bool success, bytes memory result) = _resolve(input);
        if(success) {
            return result;
        } else {
            assembly {
                revert(add(result, 32), mload(result))
            }
        }
    }
    
    function _resolve(bytes memory input) internal returns (bool, bytes memory) {
        require(input.length >= 36, "Input too short");
        bytes32 node;
        assembly {
            node := calldataload(add(input, 36))
        }
        address resolver = ens.resolver(node);
        return resolver.call(input);
    }
}
```

We should deploy this using some kind of CALL2 deployment stub that ensures the same address on all chains - need to check if there is already one in widespread use. We can then give the contract a name such as universalresolver.ens.eth.

Arachnid:
Thinking about this, we'll need to modify this in order to support wildcard resolution; the first argument will have to be the string-encoded name, which somewhat complicates things.

makoto:
https://github.com/ensdomains/ens-contracts/pull/52

jefflau:
Completed

PR #43, opened 2021-09-29
Add UniversalResolver

makoto:
Solves https://github.com/ensdomains/ens-contracts/issues/41 (except wild card)


makoto:
Rewrite using wildcard EIP https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution

PR #48, opened 2021-12-17
Rewrite resolver to use interfaces

Arachnid:


jefflau (approved):

contracts/resolvers/OwnedResolver.sol
```diff
@@ -14,7 +14,7 @@ import "./profiles/TextResolver.sol";
  * address.
  */
 contract OwnedResolver is Ownable, ABIResolver, AddrResolver, ContentHashResolver, DNSResolver, InterfaceResolver, NameResolver, PubkeyResolver, TextResolver {
-    function isAuthorised(bytes32 node) internal override view returns(bool) {
+    function isAuthorised(bytes32) internal override view returns(bool) {
```
makoto, line 17: Is the argument not to be used? If so, isn't the convention to comment out like (`/* bytes32 */`)?

makoto:
Is this purely for code tidy up or are we getting anything extra?

makoto (approved):

Arachnid:
> Is this purely for code tidy up or are we getting anything extra?

Just a desire to clean things up prior to implementing some new stuff (CCIP-read resolver etc).

PR #52, opened 2022-01-26
Write universal resolver; fix interface detection

Arachnid:
This implements a UniversalResolver, that implements ENSIP 10 to find the resolver for the name accounting for wildcards, then queries the resolver. It also supports EIP 3668; if an `OffchainLookup` is reverted, it wraps it in its own one and re-reverts, so that callbacks are passed appropriately to the resolver contract.

The PR also cleans up some stuff around how supportsInterface is implemented in the resolver contracts.


contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns(bool) {
+        return interfaceId == type(IExtendedResolver).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function callWithOffchainLookupPropagation(address target, bytes memory data, bytes4 callbackFunction) internal view returns(bytes memory ret) {
+        uint256 result;
+        uint256 size;
+        
+        assembly {
+            result := staticcall(gas(), target, add(data, 32), mload(data), 0, 0)
```
makoto, line 55: please put a comment explaining what this operation does.

contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+     * @dev Finds a resolver by recursively querying the registry, starting at the longest name and progressively
+     *      removing labels until it finds a result.
+     * @return The Resolver responsible for this name, and the namehash of the full name.
+     */
+    function findResolver(bytes calldata name) public view returns(Resolver) {
+        (address resolver,) = findResolver(name, 0);
+        return Resolver(resolver);
+    }
+
+    function findResolver(bytes calldata name, uint256 offset) internal view returns(address, bytes32) {
```
makoto, line 109: better to avoid duplicate names with l104 (maybe `findResolverWithOffset`)?

contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+contract UniversalResolver is IExtendedResolver, ERC165 {
+    using Address for address;
+
+    ENS public immutable registry;
+
+    constructor(address _registry) {
+        registry = ENS(_registry);
+    }
+
+    function resolve(bytes calldata name, bytes memory data) external override view returns(bytes memory) {
```
makoto, line 25: public/external function should have comments about the function

contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+
+        assembly {
+            returndatacopy(0, 0, size)
+            revert(0, size)
+        }
+    }
+
+    function resolveCallback(bytes calldata response, bytes calldata extraData) external view returns(bytes memory) {
```
makoto, line 94: public/external function should have comments about the function

hardhat.config.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
   solidity: {
     compilers: [
       {
-        version: "0.8.4",
+        version: "0.8.11",
```
makoto, line 73: you bumped this version to 0.8.11 but all the contracts are set to 0.8.4. is this intentional?

makoto:
Please add UniversalResolver to https://github.com/ensdomains/ens-contracts/blob/master/index.js so that the abi is exportable via npm module

test/utils/TestUniversalResolver.js
```diff
@@ -0,0 +1,87 @@
...
+            expect(result).to.equal(publicResolver.address);
+        });
+
+        it('should find a resolver on a parent name', async () => {
+            const result = await universalResolver.findResolver(dns.hexEncodeName('foo.test.eth'));
+            expect(result).to.equal(publicResolver.address);
+        });
+
+        it('should choose the resolver closest to the leaf', async () => {
+            const result = await universalResolver.findResolver(dns.hexEncodeName('sub.test.eth'));
```
makoto, line 46: Probably easier to understand if the test name is 'foo.sub.test.eth' so that it's taking from the closer parent(sub.test.eth not from test.eth) rather than the resolver set to the domain itself

contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns(bool) {
+        return interfaceId == type(IExtendedResolver).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function callWithOffchainLookupPropagation(address target, bytes memory data, bytes4 callbackFunction) internal view returns(bytes memory ret) {
+        uint256 result;
+        uint256 size;
+        
+        assembly {
```
jefflau, line 54: For the assembly code what do you think about encapsulating inside their own functions to make it easier to understand what they do

jefflau (commented):
As I mentioned via DMs, would be good to see how multicall may work in the universal resolver

contracts/utils/UniversalResolver.sol
Arachnid: I think this is clearer as-is.

hardhat.config.js
Arachnid: Yup. Contracts specify a minimum version, not a required version.

Arachnid:
@makoto @jefflau PTAL. I'm working on a batch resolution function next.

contracts/utils/UniversalResolver.sol
makoto: It's also a minor inconvenience that the frontend has to call `resolver['findResolver(bytes calldata name, uint256 offset)']` instead of just `resolver.findResolver()`. I understand that `.addr` has the similar issue but it was added due to extend the function which was unavoidable.

Arachnid: This won't affect the frontend, because the second method is internal rather than public.

makoto:
Is it possible to provide a function if the resolver of the given name supports `resolver.resolve` which I currently use to determine whether I should show wildcard related messages on our ENS manager. Wrapping the call with universal resolver makes it impossible to detect unless I call the underlying resolver contract directly which is an extra call.

contracts/utils/UniversalResolver.sol
```diff
@@ -0,0 +1,126 @@
...
+     * @param data The resolution data, as specified in ENSIP-10.
+     * @return The result of resolving the name.
+     */
+    function resolve(bytes calldata name, bytes memory data) external override view returns(bytes memory) {
+        Resolver resolver = findResolver(name);
+        if(address(resolver) == address(0)) {
+            return "";
+        }
+
+        if(resolver.supportsInterface(type(IExtendedResolver).interfaceId)) {
```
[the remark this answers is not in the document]
Arachnid: Good idea. Done!

Arachnid:
@jefflau @makoto PTAL

> Is it possible to provide a function if the resolver of the given name supports resolver.resolve which I currently use to determine whether I should show wildcard related messages on our ENS manager. Wrapping the call with universal resolver makes it impossible to detect unless I call the underlying resolver contract directly which is an extra call.

Could you use `findResolver` instead? I'd rather not make something so specific part of the interface.

jefflau (approved):
LGTM

test/utils/TestUniversalResolver.js
```diff
@@ -0,0 +1,101 @@
...
+
+    describe('resolve()', () => {
+        it('should resolve a record if `supportsInterface` throws', async () => {
+            const legacyResolver = await LegacyResolver.new();
+            await ens.setSubnodeOwner(namehash.hash('eth'), sha3('test2'), accounts[0], {from: accounts[0]});
+            await ens.setResolver(namehash.hash('test2.eth'), legacyResolver.address, {from: accounts[0]});
+            const data = (await legacyResolver.methods['addr(bytes32)'].request(namehash.hash('test.eth'))).data;
+            const result = await universalResolver.resolve(dns.hexEncodeName('test2.eth'), data);
+            const [ret] = ethers.utils.defaultAbiCoder.decode(["address"], result);
+            expect(ret).to.equal(legacyResolver.address);
```
makoto, line 63: If I read the test correctly, `address` field returns the resolver address if the resolver doesn't support wildcard (='resolve' function), but returns the address record if it does support? How do you distinguish if it is returning the resolver address or eth address record?

test/utils/TestUniversalResolver.js
```diff
@@ -0,0 +1,101 @@
...
+            const [ret] = ethers.utils.defaultAbiCoder.decode(["address"], result);
+            expect(ret).to.equal(accounts[1]);
+        });
+
+        it('should return a wrapped revert if the resolver reverts with OffchainData', async () => {
+            const data = (await publicResolver.methods["addr(bytes32)"].request(namehash.hash('offchain.test.eth'))).data;
+            // OffchainLookup(address sender, string[] urls, bytes callData, bytes4 callbackFunction, bytes extraData)
+            // This is the extraData value the universal resolver should encode
+            const extraData = ethers.utils.defaultAbiCoder.encode(['address', 'bytes4', 'bytes'], [dummyOffchainResolver.address, '0xb4a85801', data]);
+            await expect(universalResolver.resolve(dns.hexEncodeName('offchain.test.eth'), data)).to.be.revertedWith(
```
makoto, line 78: can you put `0xb4a85801` into constant?

test/utils/TestUniversalResolver.js
Arachnid: No, it always returns the result of calling `resolve` or `addr`. In this case the legacy resolver simply returns its own address.

Arachnid:
@makoto PTAL

makoto (approved):

PR #53, opened 2022-02-01
Exponential price decay contract

jefflau:
Adds a new Price Oracle contract that reduces the price by 50% per day for 28 days. Price starts at ~$100m USD and decays to 0.

* Uses a look up table to compute the correct fractional exponent
* Look up table has 16 bit resolution so there is a small error margin. 16 bit resolution gives us 1/65536 of a day, so just over 1 second per day as there are 86400 seconds per day.
* Uses bit shifting to do the non-fractional exponent. E.g. 2 days will `shiftRight()` by 2
* Negates $.37... from the starting value so we have a smooth transition to $0 after 28 days


makoto:
Please expose the abi at https://github.com/ensdomains/ens-contracts/blob/master/index.js

jefflau:
> Please expose the abi at https://github.com/ensdomains/ens-contracts/blob/master/index.js

This is done

contracts/ethregistrar/StablePriceOracle.sol
```diff
@@ -6,41 +6,51 @@ import "./StringUtils.sol";
...
     AggregatorInterface public immutable usdOracle;
 
     event OracleChanged(address oracle);
 
-    event RentPriceChanged(uint[] prices);
+    event RentPriceChanged(uint256[] prices);
 
-    bytes4 constant private INTERFACE_META_ID = bytes4(keccak256("supportsInterface(bytes4)"));
-    bytes4 constant private ORACLE_ID = bytes4(keccak256("price(string,uint256,uint256)") ^ keccak256("premium(string,uint256,uint256)"));
+    bytes4 private constant INTERFACE_META_ID =
```
Arachnid, line 27: We should use interfaceId for this.

Arachnid (changes requested):

contracts/ethregistrar/ExponentialPremiumPriceOracle.sol
```diff
@@ -0,0 +1,143 @@
...
+    uint256 constant bit14 = 917004043204671232;
+    uint256 constant bit15 = 840896415253714560;
+    uint256 constant bit16 = 707106781186547584;
+
+    /**
+     * @dev Store value in variable
+     * @param startPremium value to store
+     * @param elapsed time past since expiry
+     */
+    function decayedPremium(uint256 startPremium, uint256 elapsed)
```
Arachnid, line 43: This should probably be below _premium

contracts/ethregistrar/ExponentialPremiumPriceOracle.sol
```diff
@@ -0,0 +1,143 @@
...
+        uint256
+    ) internal view override returns (uint256) {
+        expires = expires + GRACE_PERIOD;
+        if (expires > block.timestamp) {
+            return 0;
+        }
+
+        uint256 elapsed = block.timestamp - expires;
+        uint256 premium = decayedPremium(START_PREMIUM, elapsed);
+        if (premium >= endValue) {
```
Arachnid, line 72: It shouldn't be possible for this to be false - can we just remove the if?

contracts/ethregistrar/StablePriceOracle.sol
Arachnid: Ping

Arachnid (changes requested):

contracts/ethregistrar/ExponentialPremiumPriceOracle.sol
jefflau: We don't set an end time for the premium, so the premium will keep decreasing after the last day, e.g. after day 21 it will be below $47 and so at that point, the premium would be say $40 -$47 which would be -$7 or just 0 if it saturates?

I get this when I remove it:

```
Error: VM Exception while processing transaction: reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)
```

I am checking a day past the last day in my tests to make sure we get back 0. I guess the alternative would be to just revert, but that would be almost the same as an if statement,

Arachnid (approved):

PR #54, opened 2022-02-01
Added eslint plugin babel to dev dependencies

Gh05d:


makoto:
You seem deleting lots of code as well

PR #61, opened 2022-03-15
Custom Errors. Updated docs

jefflau:


contracts/wrapper/NameWrapper.sol
```diff
@@ -162,13 +173,13 @@ contract NameWrapper is
...
-            registrant == msg.sender ||
+        if (
+            !(registrant == msg.sender ||
                 isApprovedForAll(registrant, msg.sender) ||
-                registrar.isApprovedForAll(registrant, msg.sender),
-            "NameWrapper: Sender is not owner or authorised by the owner or authorised on the .eth registrar"
-        );
+                registrar.isApprovedForAll(registrant, msg.sender))
+        ) {
+            revert UnauthorisedEthWrap(bytes32(tokenId), msg.sender);
```
Arachnid, line 181: Can we just use `Unauthorised` here?

contracts/wrapper/NameWrapper.sol
```diff
@@ -473,10 +484,9 @@ contract NameWrapper is
      */
     modifier operationAllowed(bytes32 node, uint96 fuseMask) {
         (, uint96 fuses) = getData(uint256(node));
-        require(
-            fuses & fuseMask == 0,
-            "NameWrapper: Operation prohibited by fuses"
-        );
+        if (!(fuses & fuseMask == 0)) {
```
Arachnid, line 487:
```suggestion
        if (fuses & fuseMask != 0) {
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -495,11 +505,13 @@ contract NameWrapper is
...
         (, uint96 fuses) = getData(uint256(node));
 
-        require(
-            (owner == address(0) && fuses & CANNOT_CREATE_SUBDOMAIN == 0) ||
-                (owner != address(0) && fuses & CANNOT_REPLACE_SUBDOMAIN == 0),
-            "NameWrapper: Operation prohibited by fuses"
-        );
+        if (
+            !((owner == address(0) && fuses & CANNOT_CREATE_SUBDOMAIN == 0) ||
+                (owner != address(0) && fuses & CANNOT_REPLACE_SUBDOMAIN == 0))
```
Arachnid, line 510:
```suggestion
            (owner == address(0) && fuses & CANNOT_CREATE_SUBDOMAIN != 0) ||
                (owner != address(0) && fuses & CANNOT_REPLACE_SUBDOMAIN != 0)
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -231,22 +242,23 @@ contract NameWrapper is
...
+            revert IncompatibleParent();
+        }
 
         address owner = ens.owner(node);
-        require(
-            owner == msg.sender ||
+
+        if (
+            !(owner == msg.sender ||
                 isApprovedForAll(owner, msg.sender) ||
```
Arachnid, line 257:
```suggestion
            owner != msg.sender &&
                !isApprovedForAll(owner, msg.sender) &&
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -231,22 +242,23 @@ contract NameWrapper is
...
-                ens.isApprovedForAll(owner, msg.sender),
-            "NameWrapper: Domain is not owned by the sender"
-        );
+                ens.isApprovedForAll(owner, msg.sender))
```
Arachnid, line 258:
```suggestion
                !ens.isApprovedForAll(owner, msg.sender)
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -162,13 +173,13 @@ contract NameWrapper is
     ) public override {
         uint256 tokenId = uint256(keccak256(bytes(label)));
         address registrant = registrar.ownerOf(tokenId);
-
-        require(
-            registrant == msg.sender ||
+        if (
+            !(registrant == msg.sender ||
                 isApprovedForAll(registrant, msg.sender) ||
```
Arachnid, line 178:
```suggestion
            registrant != msg.sender &&
                !isApprovedForAll(registrant, msg.sender) &&
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -162,13 +173,13 @@ contract NameWrapper is
...
-                registrar.isApprovedForAll(registrant, msg.sender),
-            "NameWrapper: Sender is not owner or authorised by the owner or authorised on the .eth registrar"
-        );
+                registrar.isApprovedForAll(registrant, msg.sender))
```
Arachnid, line 179:
```suggestion
                !registrar.isApprovedForAll(registrant, msg.sender)
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -527,10 +539,9 @@ contract NameWrapper is
         bytes calldata data
     ) public override returns (bytes4) {
         //check if it's the eth registrar ERC721
-        require(
-            msg.sender == address(registrar),
-            "NameWrapper: Wrapper only supports .eth ERC721 token transfers"
-        );
+        if (!(msg.sender == address(registrar))) {
```
Arachnid, line 542:
```suggestion
        if (msg.sender != address(registrar)) {
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -573,8 +584,12 @@ contract NameWrapper is
         pure
         returns (bytes memory ret)
     {
-        require(bytes(label).length > 0, "NameWrapper: Label too short");
-        require(bytes(label).length < 256, "NameWrapper: Label too long");
+        if (bytes(label).length < 1) {
+            revert LabelTooShort(label);
```
Arachnid, line 588: `label` here will always be the empty string, so I think we can probably leave it out as an argument.

contracts/wrapper/NameWrapper.sol
```diff
@@ -621,18 +639,19 @@ contract NameWrapper is
...
+            revert IncorrectTargetOwner(newOwner);
+        }
+
+        if (newOwner == address(this)) {
+            // Target owner cannot be the NameWrapper contract
+            revert IncorrectTargetOwner(newOwner);
+        }
+
+        if (allFusesBurned(node, CANNOT_UNWRAP)) {
+            revert OperationProhibited(node);
```
Arachnid, line 653: Perhaps we should log the fuse it was checking against too?

contracts/wrapper/NameWrapper.sol
```diff
@@ -646,10 +665,9 @@ contract NameWrapper is
         address owner,
         uint96 fuses
     ) internal override {
-        require(
-            fuses == CAN_DO_EVERYTHING || fuses & CANNOT_UNWRAP != 0,
-            "NameWrapper: Cannot burn fuses: domain can be unwrapped"
-        );
+        if (!(fuses == CAN_DO_EVERYTHING || fuses & CANNOT_UNWRAP != 0)) {
```
Arachnid, line 668:
```suggestion
        if (fuses != CAN_DO_EVERYTHING && fuses & CANNOT_UNWRAP == 0)) {
```

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
jefflau: I had it like that before, but I am logging the node. If I wanted to log it here, I would only have the labelhash. I could create the namehash here if that isn't too much overhead.

contracts/wrapper/NameWrapper.sol
jefflau: I thought about this. It's easy here to log CANNOT_UNWRAP, but less obvious for this:

(owner == address(0) && fuses & CANNOT_CREATE_SUBDOMAIN != 0) ||
	                (owner != address(0) && fuses & CANNOT_REPLACE_SUBDOMAIN != 0)

jefflau:
```
        if((owner == address(0) && fuses & CANNOT_CREATE_SUBDOMAIN != 0)){
            OperationProhibited(node, CANNOT_CREATE_SUBDOMAIN);
        }

        if((owner != address(0) && fuses & CANNOT_REPLACE_SUBDOMAIN != 0)){
            OperationProhibited(node, CANNOT_REPLACE_SUBDOMAIN);
        }

```

It's not quite right, since it's also dependent on whether the owner is 0 or not...

jefflau: Also, i'm trying to add `OperationProhibited(node, fuses)` to `ERC1155Fuse`, but it doesn't inherit from INameWrapper which has all the constants for fuses. What's the best way to get the CAN_TRANSFER fuse?

We call `_canTransfer` inside this contract instead of using the fuse directly so I don't have access it (bar hardcoding the number)

contracts/wrapper/NameWrapper.sol
Arachnid: It's just a keccak hash, so I'd say go for it.

contracts/wrapper/NameWrapper.sol
Arachnid: Good point. Let's leave it as-is.

jefflau:
This should be resolver now. I also changed labels to labelhash for consistency. PTAL @Arachnid

contracts/wrapper/NameWrapper.sol
```diff
@@ -473,10 +486,9 @@ contract NameWrapper is
      */
     modifier operationAllowed(bytes32 node, uint96 fuseMask) {
         (, uint96 fuses) = getData(uint256(node));
-        require(
-            fuses & fuseMask == 0,
-            "NameWrapper: Operation prohibited by fuses"
-        );
+        if (fuses & fuseMask != 0) {
+            revert OperationProhibited(node);
```
Arachnid, line 490: Perhaps this should log the fuses that were problematic (eg, `fuses & fuseMask`).

contracts/wrapper/NameWrapper.sol
```diff
@@ -560,21 +573,25 @@ contract NameWrapper is
...
 
     function _addLabel(string memory label, bytes memory name)
         internal
         pure
         returns (bytes memory ret)
     {
-        require(bytes(label).length > 0, "NameWrapper: Label too short");
-        require(bytes(label).length < 256, "NameWrapper: Label too long");
+        if (bytes(label).length < 1) {
+            revert LabelTooShort("");
```
Arachnid, line 590: Since this will always be the empty string, we can just leave out the argument entirely.

contracts/wrapper/NameWrapper.sol
```diff
@@ -621,18 +641,19 @@ contract NameWrapper is
...
-        );
-        require(
-            newOwner != address(this),
-            "NameWrapper: Target owner cannot be the NameWrapper contract"
-        );
-        require(
-            !allFusesBurned(node, CANNOT_UNWRAP),
-            "NameWrapper: Domain is not unwrappable"
-        );
+        if (newOwner == address(0x0)) {
```
Arachnid, line 644: These two checks can be combined into a single one.

Arachnid (approved):
A few minor comments, otherwise LGTM.

PR #64, opened 2022-03-23
Universal reverse

jefflau:


contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,104 @@
+pragma solidity ^0.8.4;
+
+import "./Strings.sol";
+
+library NameEncoder {
+    using strings for *;
+
+    function encodePart(strings.slice memory name)
```
Arachnid, line 8: This is clever, but does O(n^2) memory allocation. Which is okay for read ops, but wouldn't be great onchain.

contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,104 @@
...
+                uint8(label.len()),
+                bytes(label.toString()),
+                encodedName
+            );
+    }
+
+    function encode(string memory name) internal view returns (bytes memory) {
+        strings.slice memory sliceName = name.toSlice();
+        if (sliceName.empty()) {
+            return abi.encodePacked(uint8(0), uint8(0));
```
Arachnid, line 31: Why two terminators here?

contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,104 @@
...
+        return encodePart(sliceName);
+    }
+
+    function encodeAndHash(string memory name)
+        internal
+        view
+        returns (bytes memory, bytes32)
+    {
+        bytes memory encodedName = encode(name);
+        return (encodedName, namehash(encodedName, 0));
```
Arachnid, line 42: It'd be more efficient to produce the hash at the same time as converting the name.

contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /*
+     * @dev Returns the keccak-256 hash of a byte range.
+     * @param self The byte string to hash.
+     * @param offset The position to start hashing at.
+     * @param len The number of bytes to hash.
+     * @return The hash of the byte range.
+     */
+    function keccak(
```
Arachnid, line 52: Can we put this in strings.sol?

contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * @dev Returns the ENS namehash of a DNS-encoded name.
+     * @param self The DNS-encoded name to hash.
+     * @param offset The offset at which to start hashing.
+     * @return The namehash of the name.
+     */
+    function namehash(bytes memory self, uint256 offset)
```
Arachnid, line 69: Can we reuse this from the dnssec code?

contracts/utils/Strings.sol
```diff
@@ -0,0 +1,235 @@
+library strings {
```
Arachnid, line 1: Can you include a link to the source library and revision?

contracts/utils/UniversalResolver.sol
```diff
@@ -7,51 +7,197 @@ import "./LowLevelCallUtils.sol";
...
     using Address for address;
+    using NameEncoder for string;
 
     ENS public immutable registry;
 
     constructor(address _registry) {
         registry = ENS(_registry);
     }
 
+    function encodeName(string memory name) public view returns (bytes memory) {
```
Arachnid, line 34: Why is this function needed?

contracts/utils/UniversalResolver.sol
```diff
@@ -7,51 +7,197 @@ import "./LowLevelCallUtils.sol";
...
+        string memory resolvedName = resolver.name(reverseNamehash);
+        if (bytes(resolvedName).length == 0) {
+            return ("", new bytes[](0));
+        }
+
+        (bytes memory encodedName, bytes32 namehash) = resolvedName
+            .encodeAndHash();
+        bytes[] memory inputData = new bytes[](calls.length);
+
+        for (uint256 i = 0; i < calls.length; i++) {
```
Arachnid, line 111: Wouldn't it be more effective to have the caller supply the already-ABI-encoded calldata, and just concatenate it?

contracts/utils/UniversalResolver.sol
```diff
@@ -7,51 +7,197 @@ import "./LowLevelCallUtils.sol";
...
+                        namehash,
+                        abi.decode(data.data, (bytes4))
+                    );
+                } else {
+                    continue;
+                }
+            }
+        }
+
+        bytes memory replacedData = abi.encodeWithSignature(
```
Arachnid, line 167: Why not just call 'resolve' directly with each input?

contracts/utils/UniversalResolver.sol
```diff
@@ -7,51 +7,197 @@ import "./LowLevelCallUtils.sol";
...
+        ReverseData[] data;
+    }
+
+    /**
+     * @dev Performs ENS name reverse resolution for the supplied address and resolution data.
+     * @param name The address to resolve, in normalised and DNS-encoded form.
+     * @param calls The resolution data, as specified in ENSIP-10.
+     * @return The resolved name, and the resolved data.
+     */
+    function reverse(bytes calldata name, ReverseCall[] memory calls)
```
Arachnid, line 92: Why not pass in the namehash of the reverse node instead of the plaintext name?

contracts/utils/UniversalResolver.sol
```diff
@@ -7,51 +7,197 @@ import "./LowLevelCallUtils.sol";
...
 
-    function supportsInterface(bytes4 interfaceId) public view virtual override returns(bool) {
-        return interfaceId == type(IExtendedResolver).interfaceId || super.supportsInterface(interfaceId);
+    struct ReverseData {
+        string dataType;
+        bytes data;
+    }
+
+    struct ReverseCall {
+        string sig;
```
Arachnid, line 82: This should be the bytes4 function identifier instead of the string.

contracts/utils/UniversalResolver.sol
Arachnid: More generally, this batched approach is not going to play well with EIP 3668. It would probably make more sense to write a version of the function that does a single resolution operation, and we can wrap it in a batch function later.

Arachnid (changes requested):

contracts/utils/UniversalResolver.sol
jefflau: I guess it's not, artifact of putting it into the library and not removing it.

contracts/utils/NameEncoder.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.13;
+
+import "../wrapper/BytesUtil.sol";
+
+library NameEncoder {
+    using BytesUtils for bytes;
+
+    function encode(string memory name)
```
Arachnid, line 9: This should probably be called something like `dnsEncodeName`.

contracts/utils/UniversalResolver.sol
```diff
@@ -29,30 +39,100 @@ contract UniversalResolver is IExtendedResolver, ERC165 {
...
     }
 
-    function supportsInterface(bytes4 interfaceId) public view virtual override returns(bool) {
-        return interfaceId == type(IExtendedResolver).interfaceId || super.supportsInterface(interfaceId);
+    /**
+     * @dev Performs ENS name reverse resolution for the supplied address and resolution data.
+     * @param reverseNode The reverse node to resolve, in normalised and DNS-encoded form.
+     * @return The resolved name, and the resolved data.
+     */
+    function reverse(bytes memory reverseNode)
```
Arachnid, line 82: Can this be `calldata` instead of `memory`?

contracts/utils/UniversalResolver.sol
```diff
@@ -29,30 +39,100 @@ contract UniversalResolver is IExtendedResolver, ERC165 {
...
+        external
+        view
+        returns (string memory, bytes memory)
+    {
+        (bool reverseSuccess, bytes memory resolvedReverseData) = address(this)
```
Arachnid, line 87: Why can't this just be `this.resolve(...)`?

contracts/utils/UniversalResolver.sol
```diff
@@ -29,30 +39,100 @@ contract UniversalResolver is IExtendedResolver, ERC165 {
...
+            ),
+            address(resolver)
         );
     }
 
-    function supportsInterface(bytes4 interfaceId) public view virtual override returns(bool) {
-        return interfaceId == type(IExtendedResolver).interfaceId || super.supportsInterface(interfaceId);
+    /**
+     * @dev Performs ENS name reverse resolution for the supplied address and resolution data.
+     * @param reverseNode The reverse node to resolve, in normalised and DNS-encoded form.
```
Arachnid, line 79: This is a name, not a node. Perhaps an example here would be helpful too.

contracts/utils/UniversalResolver.sol
```diff
@@ -29,30 +39,100 @@ contract UniversalResolver is IExtendedResolver, ERC165 {
...
+        }
+
+        string memory resolvedName = abi.decode(
+            abi.decode(resolvedReverseData, (bytes)),
+            (string)
+        );
+
+        (bytes memory encodedName, bytes32 namehash) = resolvedName.encode();
+
+        (bool success, bytes memory resolvedData) = address(this).staticcall(
```
Arachnid, line 112: As above - can this just be `this.resolve` or even just `resolve`?

Arachnid (changes requested):

Arachnid (approved):

PR #67, opened 2022-04-07
Ccip-compatible multicall mixin

Arachnid:


jefflau:
@Arachnid Can we close this?

PR #69, opened 2022-04-26
Remove import statements from IABIResolver.sol

royalfork:
The `IABIResolver` interface does not require any external solidity files.  These import statements can be removed.


Arachnid (approved):

Arachnid:
Thanks!

PR #70, opened 2022-05-04
NameWrapper upgrade 

nxt3d:
Here is my PR for upgrade functionality for the NameWrapper. In the future it will be possible to deploy an upgraded NameWrapper contract, set the upgrade address to the contract, and allow all wrapped names to migrate. 

Even if the CANNOT_UNWRAP fuse is burnt for a wrapped name, it would still be possible to migrate the name to the upgraded NameWrapper contract. The fuses and resolver records of the wrapped name will be passed to the upgraded contract without changes.


contracts/wrapper/NameWrapper.sol
```diff
@@ -333,6 +362,77 @@ contract NameWrapper is
...
+
+        address resolver = ens.resolver(node);
+
+        (uint96 fuses,,) = getFuses(node);
+
+        upgradeContract.wrap(name, wrappedOwner, fuses, resolver);
+
+        // burn token and fuse data
+        _burn(uint256(node));
+    }
```
Arachnid, line 434:
```suggestion
    /**
     * @notice Upgrades a domain to a new name wrapper version.
     * @dev Can be called by the owner or an authorised caller
     * @param name The name to upgrade, in DNS format
     * @param wrappedOwner owner of the name in this contract
     */

    function upgrade(
        bytes calldata name,
        address wrappedOwner
    ) public {

        if (address(upgradeContract) == address(0)){
            revert CannotUpgrade();
        }

        (bytes32 labelhash, uint offset) = name.readLabel(0);
        bytes32 parentNode = name.namehash(offset);
        bytes32 node = _makeNode(parentNode, labelhash);

        if (!isTokenOwnerOrApproved(node, msg.sender)){
            revert Unauthorised(node, msg.sender);
        }

        address resolver = ens.resolver(node);

        (uint96 fuses,,) = getFuses(node);

        if(parentNode == ETH_NODE) {
            upgradeContract.wrapETH2LD(name[1:offset], wrappedOwner, fuses, resolver);
        } else {
            upgradeContract.wrap(name, wrappedOwner, fuses, resolver);
        }

        // burn token and fuse data
        _burn(uint256(node));
    }
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -20,6 +21,7 @@ error LabelMismatch(bytes32 labelHash, bytes32 expectedLabelhash);
 error LabelTooShort();
 error LabelTooLong(string label);
 error IncorrectTargetOwner(address owner);
+error ZeroAddress();
```
Arachnid, line 24:
```suggestion
error CannotUpgrade();
```

`ZeroAddress` doesn't explain the problem - the problem is that upgrades are not permitted at this time.

Arachnid (changes requested):

test/wrapper/NameWrapper.js
```diff
@@ -1144,7 +1153,519 @@ describe('Name Wrapper', () => {
...
+        CAN_DO_EVERYTHING,
+        EMPTY_ADDRESS
+      )
+    
+      await expect(
+        NameWrapper.upgradeETH2LD(
+        label,
+        account
+      )
+      ).to.be.reverted
```
jefflau, line 1280: Can we expect() the exact revert error

contracts/wrapper/NameWrapper.sol
```diff
@@ -20,6 +21,7 @@ error LabelMismatch(bytes32 labelHash, bytes32 expectedLabelhash);
 error LabelTooShort();
 error LabelTooLong(string label);
 error IncorrectTargetOwner(address owner);
+error CannotUpgrade();
```
jefflau, line 24: Could this be something more descriptive? If `upgrade()` reverts, I assume it probably cannot be upgraded. Maybe something like `InvalidUpgradeAddress(address address)` would be better.

test/wrapper/NameWrapper.js
```diff
@@ -1144,7 +1153,519 @@ describe('Name Wrapper', () => {
...
+      //set the upgradeContract of the NameWrapper contract
+      await NameWrapper.setUpgradeContract(NameWrapperUpgraded.address)
+
+     //make sure reclaim claimed ownership for the wrapper in registry
+      expect(await NameWrapper.upgradeContract()).to.equal(NameWrapperUpgraded.address)
+
+      //set the upgradeContract of the NameWrapper contract
+      await NameWrapper.setUpgradeContract(EMPTY_ADDRESS)
+
+      await expect(
```
jefflau, line 1306: Exact revert error

jefflau (changes requested):
We are allowing the owner and authorised operators to upgrade, so we should also test that an account that has been approved by the owner can also `upgrade()` and `upgradeETH2LD()`

nxt3d:
> 

I added two more tests

Arachnid (approved):

PR #73, opened 2022-05-18
Refactor set subnode

jefflau:


contracts/wrapper/NameWrapper.sol
```diff
@@ -657,6 +644,53 @@ contract NameWrapper is
...
+    function _setSubnodeOwnerAndWrap(
+        bytes32 parentNode,
+        string memory label,
+        address newOwner,
+        uint96 _fuses
+    ) internal {
+        bytes32 labelhash = keccak256(bytes(label));
+        bytes32 node = _makeNode(parentNode, labelhash);
+        bytes memory name = _addLabel(label, names[parentNode]);
+        ens.setSubnodeOwner(parentNode, labelhash, address(this));
```
Arachnid, line 668: This is the same as `_addLabelAndWrap` except for the call to `setSubnodeOwner`. Could that be moved into the calling function?

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
```diff
@@ -657,6 +644,53 @@ contract NameWrapper is
...
+        bytes32 labelhash,
+        address newOwner,
+        uint96 _fuses
+    ) internal {
+        bytes32 subnode = _makeNode(parentNode, labelhash);
+        (, uint96 fuses) = getData(uint256(subnode));
+
+        uint96 newFuses = fuses | _fuses;
+
+        _setData(uint256(subnode), newOwner, newFuses);
```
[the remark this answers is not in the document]
jefflau: Do you mean burnChildFuses? If so, it means we need to do the modifier checks again, and also it doesn't allow changing the owner, but this would also do the transfer to the newOwner

contracts/wrapper/NameWrapper.sol
jefflau: Good catch, I've removed this function in favour of _addLabelAndWrap

contracts/wrapper/NameWrapper.sol
```diff
@@ -657,23 +645,58 @@ contract NameWrapper is
...
     }
 
+    function _addLabelAndWrap(
+        bytes32 parentNode,
+        string memory label,
+        address newOwner,
+        uint96 _fuses
+    ) internal returns (bytes32 node, bytes32 labelhash) {
+        labelhash = keccak256(bytes(label));
+        node = _makeNode(parentNode, labelhash);
```
Arachnid, line 655: Can we pass this in as an argument to avoid calculating it again?

contracts/wrapper/NameWrapper.sol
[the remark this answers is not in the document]
Arachnid: I meant the whole set-and-transfer thing; we definitely have that functionality in other functions. Doing all these operations manually, including emitting events, seems like a code smell.

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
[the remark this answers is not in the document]
jefflau: I think this is essentially it.

We have 2 functions that do one of each:
* BurnFuses/burnChildFuses()
* safeTransferFrom()

burnFuses only burns fuses and safeTransfer does the transferring only. So this is a function that does both. I could call burnFuses and safeTransfer instead of creating this function, but it would mean we would be doing the owner checks twice.

We probably can expose this specific function publicly so it can be part of the API.

contracts/wrapper/NameWrapper.sol
jefflau: Done

contracts/wrapper/NameWrapper.sol
```diff
@@ -456,21 +416,40 @@ contract NameWrapper is
...
+    {
         bytes32 labelhash = keccak256(bytes(label));
         bytes32 node = _makeNode(parentNode, labelhash);
-        bytes memory name = _addLabel(label, names[parentNode]);
-
-        setSubnodeRecord(parentNode, labelhash, address(this), resolver, ttl);
 
-        _wrap(node, name, newOwner, _fuses);
+        if (ens.owner(node) != address(this)) {
+            ens.setSubnodeRecord(
```
Arachnid, line 435: This can go outside the if statement, since it's the same in both branches.

contracts/wrapper/NameWrapper.sol
```diff
@@ -431,19 +384,26 @@ contract NameWrapper is
      * @param _fuses initial fuses for the wrapped subdomain
      */
 
-    function setSubnodeOwnerAndWrap(
+    function setSubnodeOwner(
```
Arachnid, line 387: Maybe we can delete this entirely? The only thing it does differently to `setSubnodeRecord` is set resolver and ttl; is it acceptable to just offer that interface alone?

contracts/wrapper/NameWrapper.sol
```diff
@@ -675,23 +654,47 @@ contract NameWrapper is
...
+        _wrap(node, name, newOwner, _fuses);
+    }
+
+    function _transferAndBurnFuses(
+        bytes32 node,
+        address newOwner,
+        uint96 _fuses
+    ) internal {
+        (address owner, ) = getData(uint256(node));
+        _transfer(owner, newOwner, uint256(node), 1, "");
```
Arachnid, line 674: Probably worth adding a conditional here so it doesn't do this if `owner == newOwner`.

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
jefflau: Hmm, it does take less gas though. I think it's acceptable to just offer that alone since it does everything the other function does. 

If you didn't want to set the resolver though, it would be quite a bit cheaper to use `setSubnodeOwner`.

Arachnid: Fair enough. Let's leave it as-is.

contracts/wrapper/NameWrapper.sol
jefflau: Can't do this, because it needs to go before wrap for re-entrancy, which I think we discussed today.

contracts/wrapper/NameWrapper.sol
```diff
@@ -495,7 +474,11 @@ contract NameWrapper is
             CANNOT_TRANSFER | CANNOT_SET_RESOLVER | CANNOT_SET_TTL
         )
     {
-        ens.setRecord(node, owner, resolver, ttl);
+        ens.setRecord(node, address(this), resolver, ttl);
+        (address oldOwner, ) = getData(uint256(node));
+        if (owner != oldOwner){
```
Arachnid, line 479: Maybe this check should go in `_transfer` instead?

Arachnid (approved):

PR #77, opened 2022-05-24
Subdomain registrar

jefflau:
* Allows burning of PARENT_CONTROL fuse
* Allows setting fees
* Allows expiration of subdomains
* Allows setting records at the same time as registering subdomains


Arachnid (changes requested):
It might be worth considering making this contract ERC20-native. With approvals and `transferFrom`, payments could be made directly to the beneficiary, and name owners could set prices in any token they want.

jefflau:
> It might be worth considering making this contract ERC20-native. With approvals and `transferFrom`, payments could be made directly to the beneficiary, and name owners could set prices in any token they want.

I think this is a good idea

PR #86, opened 2022-06-17
Add deploy scripts

TateB:
- Adds deploy scripts for full ENS contract suite
- Adds callbacks for running other scripts before root contract and before new ETHRegistrarController
- Adds archived deployments for when a deployment is replaced/overridden
- Switched to typescript for all deploy scripts and hardhat
- Added deploy scripts to the NPM export


hardhat.config.ts
```diff
@@ -0,0 +1,142 @@
...
+
+task('accounts', 'Prints the list of accounts', async (_, hre) => {
+  const accounts = await hre.ethers.getSigners()
+
+  for (const account of accounts) {
+    console.log(account.address)
+  }
+})
+
+task('save', 'Saves a specified contract as a deployed contract')
```
Arachnid, line 29: Wouldn't it make more sense to make this command save *all* the contracts from the current deployment files, if they've been modified since the previous snapshot? Then you don't need to supply any arguments and are less likely to miss something.

hardhat.config.ts
```diff
@@ -0,0 +1,142 @@
...
+      const archivePath = `${archivedDeploymentPath}/${archiveName}.sol`
+
+      if (existsSync(archivePath)) {
+        throw new Error('Archive already exists')
+      }
+
+      let newArtifact = artifact
+      newArtifact.contractName = archiveName
+      newArtifact.sourceName = archivePath.substring(2)
+
```
Arachnid, line 47: Maybe log the commit/tree hash here too?

Arachnid (changes requested):
Can you expand a bit on the use-case here? How will the archived deployment files be used/accessed?

hardhat.config.ts
TateB: Since it's mostly meant to be run alongside a deployment, saving all the contracts probably isn't necessary since the archive is mostly used for the bytecode and ABI of the given deployed contract. If you were to save all the contracts, the snapshot might not be accurate with what is actually deployed on the network.

hardhat.config.ts
TateB: Could definitely do this, although the artifacts follow the hardhat-deploy types so adding it would be not following their spec

TateB:
> Can you expand a bit on the use-case here? How will the archived deployment files be used/accessed?

The archives allow previous contract versions to be used in deployments without needing to reference previous versions of the repo. Currently they are used to deploy the previous version of the ETHRegistrarController and PublicResolver. For the ETHRegistrarController, this allows unwrapped names to be registered in a local environment easily, meaning that unwrapped names can easily be registered and tested. For the PublicResolver, this allows the previous resolver to be used, meaning that compatibility checks can be tested. There are also probably more use-cases we will come into in the future.

This _could_ all be done in a separate repo (which it currently is), but doing it within ens-contracts provides a centralised source of deployment scripts so no complex dependency ordering is needed, as well as allowing someone to update the contracts and associated deployment scripts, and easily have that update used within ens-app testing and other ecosystem repos.

Alongside all this, providing the archives in the NPM package allows for an easily accessible source of accurate previous contract deployments, if someone might need it for their own testing or similar.

hardhat.config.ts
Arachnid: Right, but the goal is to have an accurate snapshot of past versions of contracts, correct? Rather than asking people to remember to save individual contracts as they're changed, we should probably have a function that just makes sure there's a snapshot of the current version of each contract stored, and that can be run idempotently.

hardhat.config.ts
Arachnid: If it will let you extend the spec, it would be good to have that data.

TateB:
> How will these artifacts be used for deployments? Could that code be included in the repo too?

It will be used for test deployments. Currently it'll be used in App v3 and ENSjs v3, but anyone else would also be able to if needed.

What code are you referencing?

Arachnid:
> What code are you referencing?

Whatever code is needed to take the artifact files and deploy them. I assume the standard deployments in this repo can't be used, since they work with the latest version of each contract.

Arachnid:
Sounds good then. After you make the other changes I mentioned it LGTM.

hardhat.config.ts
TateB: The goal is to have an accurate snapshot of previous deployed contracts, if a contract isn't actually deployed from the snapshot I don't think it's particularly useful.

Arachnid (approved):

PR #94, opened 2022-06-30
Upgrade namewrapper

jefflau:


contracts/wrapper/NameWrapper.sol
```diff
@@ -363,7 +391,62 @@ contract NameWrapper is
...
+     * @param resolver resolver contract for this name
+     */
+
+    function upgrade(
+        bytes32 parentNode,
+        string calldata label,
+        address wrappedOwner,
+        address resolver
+    ) public {
+        bytes32 labelhash = keccak256(bytes(label));
```
Arachnid, line 435: Doesn't this function need to check that the parent name has been upgraded first?

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
jefflau: Decided we're not going to do this in favour of having checks in the new wrapper to reduce gas on upgrading. Added notes on upgrading in the README

PR #101, opened 2022-07-06
Tiny gas optimization in StablePriceOracle.sol

0xc0de4c0ffee:
As there'll always be more domains with >=5 chars, 
changing oracle price logic to check >=5 chars first will save tiny (~59) gas during registration and renewal per domain.

Domains with <5 chars are already paying extra for registration so few gWei shouldn't matter to them.


Arachnid (approved):
Thanks!

PR #103, opened 2022-07-07
Feature/generate index js

TateB:
Created a script that generates the index.js file based on the built contract ABIs, and added to the build script.


jefflau:
This adds every Solidity contract into index.js, but I'm not sure if we want to expose every contract, such as pure testing contracts:

```
const TestNameEncoder = require('./build/contracts/TestNameEncoder')
const TestRRUtils = require('./build/contracts/TestRRUtils')
const TestRegistrar = require('./build/contracts/TestRegistrar')
const TestResolver = require('./build/contracts/TestResolver')
const TextResolver = require('./build/contracts/TextResolver')
const UniversalResolver = require('./build/contracts/UniversalResolver')
const UpgradedNameWrapperMock = require('./build/contracts/UpgradedNameWrapperMock')
```
The ones with Mock/Test probably don't need to be exported and probably can be omitted from the npm package to save on size.

Also `NameGriefer.sol` is an example contract to test for a vulnerability on the NameWrapper. I guess we could rename it to `Test` and regex out anything that starts with `Test`

PR #110, opened 2022-07-26
add opensea and etherscan indexed function and variable

Specialist-K:
Hello, ens team,

We have noticed display issues on NFT marketplaces and etherscan that could be fixed by adding a few variables.

image

The unidentified contract can be fixed by adding a variable "name" in the smart contract (maybe inside the erc1155Fuse.sol file) that the user could set up using the contract at contract deployment.
image

The max Total supply issue on Opensea can be fixed by adding a function named "totalSupply()"
image

Thank you for your time.


jefflau:
Hi sorry for the very slow reply to this.

We will add the name to the contract, but we won't be adding the total supply as it adds additional overhead to the contract for limited gain. I'll open a new PR for this instead editing this PR, so I'll close this. Thanks for the original PR.

PR #114, opened 2022-08-03
remove getFuses

nxt3d:


jefflau:
I remember adding 'reverted with an unrecognized . . .', before. Do you know why we need to add this back in now?

nxt3d:
> I remember adding 'reverted with an unrecognized . . .', before. Do you know why we need to add this back in now?

image

Arachnid (approved):

PR #115, opened 2022-08-12
add hh seed action, fix broken flow

mdtanrikulu:


TateB:
can the old scripts be removed?

TateB (approved):

PR #121, opened 2022-09-04
Add `_maxCommitmentAge` check in `constructor` to ensure `_consumeCommitment` is working.

### LINKED ISSUE #120: `_consumeCommitment` in `ETHRegistrarController` need logic improvement
```solidity
function _consumeCommitment(
        string memory name,
        uint256 duration,
        bytes32 commitment
    ) internal {
        // Require a valid commitment (is old enough and is committed)
        require(
            commitments[commitment] + minCommitmentAge <= block.timestamp,
            "ETHRegistrarController: Commitment is not valid"
        );

        // If the commitment is too old, or the name is registered, stop
        require(
            commitments[commitment] + maxCommitmentAge > block.timestamp,
            "ETHRegistrarController: Commitment has expired"
        );
.....
```
Apparently, the first require assertion can not check whether the commitment is `committed` or not because the solidity map ping supposes that every `key` of the mapping exists with the default zero value.
Luckily, the `maxCommitmentAge` in the second require assertion is small enough so that if a commitment is not `committed` it will always throw the error.
If other projects forked ens and initialized `maxCommitmentAge` with a very big value, then the `_consumeCommitment` validation is escaped.

adu-web3:
close #120

### DIFF
--- a/contracts/ethregistrar/ETHRegistrarController.sol
+++ b/contracts/ethregistrar/ETHRegistrarController.sol
@@ -55,6 +55,7 @@ contract ETHRegistrarController is Ownable, IETHRegistrarController {
         INameWrapper _nameWrapper
     ) {
         require(_maxCommitmentAge > _minCommitmentAge);
+        require(_maxCommitmentAge < block.timestamp);
 
         base = _base;
         prices = _prices;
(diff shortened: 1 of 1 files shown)


Arachnid:
Your analysis is incorrect; a large `maxCommitmentAge` will not result in an error here.

PR #125, opened 2022-09-19
Remove ETH_NODE from ETHRegistrarController

ququzone:


jefflau:
Thanks for this, however we do not want to called baseNode() as it actually costs us another 2000 gas for calling another contract, which is why we have hardcoded `ETH_NODE`

PR #126, opened 2022-09-20
Feature/versionable resolvers

TateB:
- Versionable resolvers allows all resolver records to be derived from a `recordVersion` 
- This means that records can be efficiently completely cleared, while still maintaining full compatibility with existing resolver infrastructure.
- `recordVersion` can also be set to any number, which allows reverting versions, and even switching between different sets of records.


contracts/resolvers/profiles/DNSResolver.sol
```diff
@@ -18,24 +18,24 @@ abstract contract DNSResolver is
...
     // resource containing a single zonefile.
     // node => contenthash
-    mapping(bytes32 => bytes) private zonehashes;
+    mapping(uint64 => mapping(bytes32 => bytes)) private versionable_zonehashes;
 
     // Version the mapping for each zone.  This allows users who have lost
     // track of their entries to effectively delete an entire zone by bumping
     // the version number.
     // node => version
-    mapping(bytes32 => uint256) private versions;
```
Arachnid, line 27: Can you replace the version mechanism here with the new one, rather than stacking them?

Arachnid (changes requested):

Arachnid (approved):
Approved, but let's think about how to handle changes to the repo and tagging the live version before we merge.

PR #128, opened 2022-09-20
ETHRegisterControllerV2

alextnetto:
### New Features: 

- Referral for Reveal and Renew domains.
- Batch Commit, Reveal, and Renew.
- Payment in the commit stage with extra tips to attract the registrar Dapp to reveal the name(s), covering gas costs.

You can read more about all we did in [this medium article](https://medium.com/blockful/ethregistercontrollerv2-29d04eec48a5)

Since we changed a lot from the previous version, we created a V2 for the new files

- ETHRegistrarControllerV2.sol
- IETHRegistrarControllerV2.sol
- TestEthRegistrarControllerV2.sol

These features will reward front-ends (registrar Dapps), increase domain registration, and enlarge the community, creating third-party services for the ENS ecosystem.

By implementing the referrer as an option in the register and renew functions, the protocol will provide commission when executing the service and share a fee previously settled by the DAO, with the referrer.

Front-end Dapps will be built to profit on the refer share. And furthermore can be done with the payment with commitment, which can incentivize a relation of trust between the final user and front-end Dapp, enabling names to be purchased with a better UX, by a single transaction.

To save even more in gas costs, the batch feature will vastly increase the intention of acquiring new names, since the gas cost will rapidly decay when purchasing more than 1 name.


lcfr-eth:
Has this passed a DAO vote yet? 
The last time these changes were mentioned on the forums I asked this and was told it would still require a DAO vote. 

While the functionalities may have benefit for some. 

Personally I cant see how referral fees help anyone but influencers with huge followings - It may even be used to scam said followers and exploit/game the referral system in other ways.

I think frontends are capable of setting their own "referral fees" much easier by just tacking it on to the msg.value based on their wants how we do at ens.vision - this allows a competitiveness among different frontend builds as well creating / finding a true market value of such "fees"

It feels like it's adding unnecessary complications to the core contracts which could be abstracted by third parties who are interested in these features them selves and gives third parties something to build.  

For these reasons I think you should organize a proper DAO vote on the individual features / changes you want to make before attempting to merge to main.

Alextnetto:
> Has this passed a DAO vote yet? The last time these changes were mentioned on the forums I asked this and was told it would still require a DAO vote.

https://www.ensgrants.xyz/rounds/1/proposals/16

Some other benefits are the gas savings in these new features as shown [here](https://docs.google.com/spreadsheets/d/18bcSbpRuz2cPuy2_foI3jQLFJS_WYHciOGgEWkITtnI/edit#gid=0)

Arachnid:
Outstanding work! I'll give this a more detailed review soon, but I had one initial thought about the tips/referral mechanism that bears thinking about. What if, instead of maintaining a mapping to individual commits for pre-payment, we keep a per-account balance that can be used both to fund registrations and for referral fees? A commit-with-fee can fund this account, and referral fees can be added to the users' balance, while a reveal consumes any balance in the user's account, sending the remainder back to them so they don't have to manage withdrawals.

This will require keeping a separate counter of accumulated fees that can be withdrawn to the DAO, but I think it should still work out lower gas-cost than sending the referrer their fee in each TX, and simplifies the UX and the code.

lcfr-eth:
> > Has this passed a DAO vote yet? The last time these changes were mentioned on the forums I asked this and was told it would still require a DAO vote.
> 
> https://www.ensgrants.xyz/rounds/1/proposals/16
> 
> Some other benefits are the gas savings in these new features as shown [here](https://docs.google.com/spreadsheets/d/18bcSbpRuz2cPuy2_foI3jQLFJS_WYHciOGgEWkITtnI/edit#gid=0)

This is not a DAO wide vote though?
This is a small grant. Which you were awarded to research and develop. 

As far as I'm aware @Arachnid core contract / protocol changes still require an official DAO wide vote? 

It almost seems like trying to side-step the DAO by attempting to merge this here without bringing it to the DAO forum and putting forward a full vote via snapshot etc.

jefflau:
Awesome stuff!

There are a few changes we made post audit, so this will need to be merged into your V2 code as well. One example is:

https://github.com/ensdomains/ens-contracts/blob/master/contracts/ethregistrar/ETHRegistrarController.sol#L308

Please have a look at the current .eth controller contract for other additional changes. I will also find some time to take a closer look

contracts/ethregistrar/ETHRegistrarControllerV2.sol
```diff
@@ -0,0 +1,390 @@
...
+            interfaceID == type(IERC165).interfaceId ||
+            interfaceID == type(IETHRegistrarControllerV2).interfaceId;
+    }
+
+    /**
+     * @dev Sets the new referral fee in percentage with one decimal.
+     * @param _referralFee The fee to be used. From 0 to 1000.
+     */
+    function setReferralFee(uint256 _referralFee) external onlyOwner {
+        require(
```
axic, line 121: I'd suggest to use custom errors as opposed to text.

alexvansande:
@Arachnid "What if, instead of maintaining a mapping to individual commits for pre-payment, we keep a per-account balance that can be used both to fund registrations and for referral fees?"

I agree. This is actually something I have disagreed on implementation with @Alextnetto. Initially we had agreed to have a general balance for everything: fees, referrals, tips, funds, etc, but in the end the team argued that it was actually more costly in terms of gas to save the new balance than just to [send the ether directly](https://github.com/ensdomains/ens-contracts/pull/128/files#diff-466c754e608ca49f1bdb936ed141c22fc44baa9b5e222849053d49d88d300615R333). As they were optimizing for gas I let it pass, but I'm still curious about why that would be the case.

Alextnetto:
> Outstanding work! I'll give this a more detailed review soon, but I had one initial thought about the tips/referral mechanism that bears thinking about. What if, instead of maintaining a mapping to individual commits for pre-payment, we keep a per-account balance that can be used both to fund registrations and for referral fees? 

[Here](https://docs.google.com/document/d/1-nmFDOm1XU9fzpNNH7HSlCD-h1PiJ4TUhGn9SFmuN-s/edit) is why we send the ether directly

Arachnid:
> [Here](https://docs.google.com/document/d/1-nmFDOm1XU9fzpNNH7HSlCD-h1PiJ4TUhGn9SFmuN-s/edit) is why we send the ether directly

Did you test this in the case where the accounts already have balances? Setting a storage slot from zero to nonzero costs 20k gas, but changing its value if it's already nonzero costs only 5k.

Alextnetto:
Heyy friENS, what's missing for getting this for a voting on DAO, can I do something to help, review, modify?

PR #143, opened 2022-10-26
remove repetitive code

lcfr-eth:
ens.setSubnodeRecord called above if/else vs in both cases


contracts/wrapper/NameWrapper.sol
```diff
@@ -579,23 +579,17 @@ contract NameWrapper is
         bytes32 labelhash = keccak256(bytes(label));
         node = _makeNode(parentNode, labelhash);
         expiry = _checkParentFusesAndExpiry(parentNode, node, fuses, expiry);
+        
+        ens.setSubnodeRecord(
```
jefflau, line 583: We can't do this, because it messes with the `isWrapped()` check

jefflau: The only solution is to have a variable and then we could just write the function once, but we're out of stack slots

PR #145, opened 2022-10-28
Wrapper expiry refactor

jefflau:
- [x] .eth fuse
- [x] Allow .eth fuse to be burned without PCC or CU
- [x] Protect again .eth fuse being burned outside _wrapETH2LD
    - [x] setSubnodeOwner
    - [x] setSubnodeRecord
    - [x] setChildFuses
    - [x] Tests for all 3
- [x] split fuses between parent controlled and child controlled (16/16)
- [x] revert if trying to set last 16bits fuses for setFuses
    - [x] Test for this
- [x] Revert if trying to set last 16 bits of _wrapETH2LD()
    - [x] Test for this
- [x] hasExpiry name fuse.  (probably PCC anyway)
- [x] 1 expiry for .eth and subdomains
- [x] Only set owner to 0 if PCC was burned
- [x] Expired names can’t set fuses so thats fine
- [x] think about how grace period is affected - use renew?
- [x] Test for .eth names expiring
    - [x] safeTransferFrom
    - [x] UnwrapETH2LD()
    - [x] setFuses()
- [x] Test for subnames expiring, cannot unwrap
    - [x] unwrap()
    - [x] setFuses()
- [x] Allow TLD owners to setChildFuses. parentNode == ROOT_NODE. check registry owner
    - [x] Test for TLD owners to setChildFues
- [x] setRecord() setting record to 0 unwraps
    - [x] Test that setting record to 0 unwraps
- [x] Figure out if _mint needs to have OR’d fuses
- [x] Allow parent fuses to be burnt without having to burn PCC
    - [x] Test parent fuses can be burnt
    - [x] Test for parent fuses are not allowed to be burnt after PCC is burnt
- [x] Add test for Approved user/contract


contracts/wrapper/NameWrapper.sol
```diff
@@ -41,11 +39,15 @@ contract NameWrapper is
...
     bytes32 private constant ETH_NODE =
         0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae;
+    bytes32 private constant ETH_LABELHASH =
+        0x4f5b812789fc606be1b3b16908db13fc7a9adf7ca72641f84d75b47069d3d7f0;
     bytes32 private constant ROOT_NODE =
         0x0000000000000000000000000000000000000000000000000000000000000000;
 
     INameWrapperUpgrade public upgradeContract;
     uint64 private constant MAX_EXPIRY = type(uint64).max;
+    uint32 private constant PARENT_CONTROLLED_FUSES =
```
Arachnid, line 49: I think this is probably clearer as just a constant.

Arachnid, line 49: I think this is probably clearer as just a constant.

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
```diff
@@ -1005,22 +1026,30 @@ contract NameWrapper is
...
-        if (fuses & PARENT_CANNOT_CONTROL != 0) {
-            // Only the parent can burn the PARENT_CANNOT_CONTROL fuse.
+        // Only the parent can burn the parent controlled fuses.
+        if (fuses & PARENT_CONTROLLED_FUSES != 0) {
             revert Unauthorised(node, msg.sender);
         }
     }
+
+    function _checkDotEthFuse(bytes32 node, uint32 fuses) internal pure {
+        if (fuses & IS_DOT_ETH == IS_DOT_ETH) {
```
makoto, line 1050: is this equivalent of `if(fuses == 0)`?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -215,37 +210,34 @@ contract ETHRegistrarController is
         payable
         override
     {
-        _renew(name, duration, 0, 0);
+        _renew(name, duration, 0);
     }
 
     function renewWithFuses(
```
makoto, line 216: Do we still need two versions (renew with fuse 0 and renew with fuses)? Can we now just always renew with the same fuse they had earlier?

contracts/wrapper/NameWrapper.sol
```diff
@@ -209,7 +228,6 @@ contract NameWrapper is
      * @param label Label as a string of the .eth domain to wrap
      * @param wrappedOwner Owner of the name in this contract
      * @param fuses Initial fuses to set
-     * @param expiry When the fuses will expire
      * @param resolver Resolver contract address
      * @return Normalised expiry of when the fuses expire
```
makoto, line 232: Do we still return normalised expiry?

contracts/wrapper/NameWrapper.sol
```diff
@@ -942,29 +964,28 @@ contract NameWrapper is
...
         address resolver
     ) private returns (uint64) {
         // Mint a new ERC1155 token with fuses
         // Set PARENT_CANNOT_REPLACE to reflect wrapper + registrar control over the 2LD
         bytes32 labelhash = keccak256(bytes(label));
         bytes32 node = _makeNode(ETH_NODE, labelhash);
-        uint32 oldFuses;
+        _checkForParentControlledFuses(node, fuses);
 
-        (, oldFuses, expiry) = _getETH2LDDataAndNormaliseExpiry(
```
makoto, line 954: Please remove `function _getETH2LDDataAndNormaliseExpiry` on line 924 which is no longer in use by anyone (and probably bunch of tests as well

contracts/wrapper/NameWrapper.sol
```diff
@@ -622,9 +632,14 @@ contract NameWrapper is
             CANNOT_TRANSFER | CANNOT_SET_RESOLVER | CANNOT_SET_TTL
         )
     {
-        ens.setRecord(node, address(this), resolver, ttl);
-        (address oldOwner, , ) = getData(uint256(node));
-        _transfer(oldOwner, owner, uint256(node), 1, "");
+        if (owner == address(0)) {
```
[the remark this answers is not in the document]
jefflau: It's the same behavior as `setSubnodeOwner()`, which unwraps when owner is set to 0. To be consistent, I added this behavior to setRecord so you could unwrap and burn a name but setting it to 0. I don't think it creates any bugs if we remove it, but it just means that you can have a wrapped name where the owner is 0. For `setSubnodeOwner()` this is definitely necessary, because settings the owner to 0 is how define deleting a subname. In this situation it would be effectively the same, but using the owner to setResolver/TTL and then unwrap, instead of calling `setRecord()` and then calling `unwrap()`.

The options I see are:

1) Leave as-is, owner can be burned to 0 with the NameWrapper still as the owner.
2) Implement the above behavior. Setting to 0, burns token (unwraps) and then sets the owner to 0 in the registry as if the domain has been deleted. Actually thinking about this now, I haven't accounted for .eth names, which would not be unwrapped.
3) Revert when setting the owner to 0

contracts/wrapper/NameWrapper.sol
```diff
@@ -1005,22 +1026,30 @@ contract NameWrapper is
...
     {
-        if (fuses & PARENT_CANNOT_CONTROL != 0) {
-            // Only the parent can burn the PARENT_CANNOT_CONTROL fuse.
+        // Only the parent can burn the parent controlled fuses.
+        if (fuses & PARENT_CONTROLLED_FUSES != 0) {
             revert Unauthorised(node, msg.sender);
         }
     }
+
+    function _checkDotEthFuse(bytes32 node, uint32 fuses) internal pure {
```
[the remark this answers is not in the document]
jefflau: I'm not sure that would work in this situation. `IS_DOT_ETH` should not be settable at all by anyone otherwise someone could spoof a domain being .eth and access logic that is only meant for .eth names

contracts/wrapper/NameWrapper.sol
jefflau: Will change

contracts/wrapper/INameWrapper.sol
```diff
@@ -12,7 +12,9 @@ uint32 constant CANNOT_TRANSFER = 4;
 uint32 constant CANNOT_SET_RESOLVER = 8;
 uint32 constant CANNOT_SET_TTL = 16;
 uint32 constant CANNOT_CREATE_SUBDOMAIN = 32;
-uint32 constant PARENT_CANNOT_CONTROL = 64;
+//uint16 reserved for parent controlled fuses
+uint32 constant PARENT_CANNOT_CONTROL = 2**16;
```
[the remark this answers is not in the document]
jefflau: Will change

contracts/wrapper/NameWrapper.sol
jefflau: No, it is saying in plain english "Is `IS_DOT_ETH` fuse contained inside the variable `fuses`"

contracts/ethregistrar/ETHRegistrarController.sol
jefflau: No, we need both. Because non-owners should not be able to burn fuses at all. So the second function did two things

1. Burn any additional fuses they want
2. Extend expiry

Now they can just do 1., however we could force them to just do a second transaction.

contracts/wrapper/NameWrapper.sol
jefflau: We just return the expiry now. I guess it could be removed since it can't really be normalised, but doesn't seem like a bad thing to return since when you wrap, you don't immediately have that information to hand. You'd have to call another function.

contracts/wrapper/NameWrapper.sol
jefflau: Good catch, will remove and check tests for normalisation of .eth names

contracts/wrapper/NameWrapper.sol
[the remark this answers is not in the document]
Arachnid: I think `setSubnodeOwner` with an owner of 0 is the only way for a name owner to delete subdomains, right? So we can't eliminate that behaviour. I think you're right that this is the correct behaviour here too for consistency - though as you observe it also has to handle .eth 2LDs.

If we're going to start adding special code for .eth 2LDs that decodes the first label, is it viable for us to do away with all the special methods that work exclusively on .eth 2LDs and do that everywhere?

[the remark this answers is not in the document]
jefflau: That's interesting. I will have a think and explore that today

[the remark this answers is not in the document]
jefflau: Actually my first thought is: it only really works for these functions *because* we have the `IS_DOT_ETH` fuse. And that fuse is set when a name is wrapped. If we did away with `wrapETH2LD()` and had it use the `wrap` method instead. We'd have to do the string manipulation in solidity that increases gas by quite a bit (which you argued against), which is why we have the `IS_DOT_ETH` fuse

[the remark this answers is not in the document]
Arachnid: What about `unwrapETH2LD` and `upgradeETH2LD`?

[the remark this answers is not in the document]
jefflau: I checked these. They both take parentNode and so there was nothing stopping us merging the logic together. It would just be an if statement and splitting the logic. I'm not sure it's worthwhile changing this logic given we made an explicit decision to split them in the past. For example:

`upgradeETH2LD` would call `wrapETH2LD `and upgrade would call `setSubnodeRecord`.

For `unwrapETH2LD` i don't think any of the current changes have changed any logic or made it simpler,.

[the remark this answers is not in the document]
jefflau: As discussed I've changed `setRecord()` to unwrap when set to 0 for subdomains or DNS names, and revert for .eth since we can't transfer to 0 for .eth

contracts/wrapper/NameWrapper.sol
```diff
@@ -479,26 +490,20 @@ contract NameWrapper is
         uint64 expiry
```
Arachnid, line 490: The docstring needs updating here.

Arachnid (changes requested):

PR #149, opened 2022-11-14
Switch ownerOf to expire and getData to return raw owner data

jefflau:


contracts/wrapper/NameWrapper.sol
```diff
@@ -302,6 +314,7 @@ contract NameWrapper is
             revert RenewalTooShort(bytes32(tokenId));
         }
         if (isWrapped(node)) {
+            // renew() protected by BaseRegistrar for non-owners, so owner should be correct owner
```
Arachnid, line 317: What does this mean?

contracts/wrapper/NameWrapper.sol
```diff
@@ -412,6 +425,7 @@ contract NameWrapper is
         operationAllowed(node, CANNOT_BURN_FUSES)
         returns (uint32)
     {
+        // owner protected by onlyTokenOwner
```
Arachnid, line 428: I think these comments should be removed.

contracts/wrapper/NameWrapper.sol
```diff
@@ -816,6 +832,7 @@ contract NameWrapper is
         uint64 expiry
     ) internal override {
         _canFusesBeBurned(node, fuses);
+        // will burn token even if the token is expired - probably good to show the flow of the token?
```
Arachnid, line 835: Yes, this is the correct behaviour.

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
```diff
@@ -100,9 +100,24 @@ contract NameWrapper is
         public
         view
         override(ERC1155Fuse, INameWrapper)
-        returns (address owner)
+        returns (address)
     {
-        return super.ownerOf(id);
+        (address owner, uint32 fuses, uint64 expiry) = super.getData(id);
```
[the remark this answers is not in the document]
jefflau: getData zeros out fuses on expiry, so i can’t properly make the second check

contracts/wrapper/NameWrapper.sol
jefflau: They were just comments to note there was no way to accidentally set the owner to the previous owner when they don’t own it. Will remove the comments once approved

contracts/wrapper/NameWrapper.sol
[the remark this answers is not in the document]
Arachnid: Can we make `getData` return the fuses as-is?

contracts/wrapper/NameWrapper.sol
```diff
@@ -491,11 +499,21 @@ contract NameWrapper is
...
             uint256(node)
         );
+
+        if (oldExpiry < block.timestamp) {
+            oldFuses = 0;
+        }
         // max expiry is set to the expiry of the parent
         (, uint32 parentFuses, uint64 maxExpiry) = getData(uint256(parentNode));
+
+        if (maxExpiry < block.timestamp) {
```
Arachnid, line 513: I don't think this is necessary, because if `maxExpiry` has already passed, you can't set fuses on the child name anyway.

contracts/wrapper/NameWrapper.sol
```diff
@@ -517,6 +535,11 @@ contract NameWrapper is
         ) {
             revert OperationProhibited(node);
         }
+
+        if (expiry < block.timestamp) {
```
Arachnid, line 539: I don't think this is necessary either, because anything reading the fuses will treat them as 0 if the name is expired.

contracts/wrapper/NameWrapper.sol
```diff
@@ -682,7 +706,10 @@ contract NameWrapper is
      */
 
     modifier operationAllowed(bytes32 node, uint32 fuseMask) {
-        (, uint32 fuses, ) = getData(uint256(node));
+        (, uint32 fuses, uint64 expiry) = getData(uint256(node));
+        if (expiry < block.timestamp) {
+            fuses = 0;
```
Arachnid, line 711: You can simply return true here.

contracts/wrapper/NameWrapper.sol
```diff
@@ -704,12 +731,20 @@ contract NameWrapper is
         address owner = ens.owner(subnode);
 
         if (owner == address(0)) {
-            (, uint32 fuses, ) = getData(uint256(node));
+            (, uint32 fuses, uint64 expiry) = getData(uint256(node));
+            if (expiry < block.timestamp) {
```
Arachnid, line 735: You can combine this with the following `if` statement.

contracts/wrapper/NameWrapper.sol
```diff
@@ -704,12 +731,20 @@ contract NameWrapper is
...
+            }
             if (fuses & CANNOT_CREATE_SUBDOMAIN != 0) {
                 revert OperationProhibited(subnode);
             }
         } else {
-            (, uint32 subnodeFuses, ) = getData(uint256(subnode));
+            (, uint32 subnodeFuses, uint64 subnodeExpiry) = getData(
+                uint256(subnode)
+            );
+            if (subnodeExpiry < block.timestamp) {
```
Arachnid, line 745: You can combine this with the following `if` statement.

contracts/wrapper/NameWrapper.sol
```diff
@@ -731,7 +766,10 @@ contract NameWrapper is
         override
         returns (bool)
     {
-        (, uint32 fuses, ) = getData(uint256(node));
+        (, uint32 fuses, uint64 expiry) = getData(uint256(node));
+        if (expiry < block.timestamp) {
```
Arachnid, line 770: You can combine this with the logic in the return statement.

contracts/wrapper/NameWrapper.sol
```diff
@@ -888,6 +928,9 @@ contract NameWrapper is
         if (names[node].length == 0) {
             names[node] = name;
         }
+        if (expiry < block.timestamp) {
```
Arachnid, line 931: I don't think this is needed, as setting expired fuses is harmless.

contracts/wrapper/NameWrapper.sol
```diff
@@ -905,6 +948,9 @@ contract NameWrapper is
     ) internal view returns (uint64) {
         (, , uint64 oldExpiry) = getData(uint256(node));
         (, uint32 parentFuses, uint64 maxExpiry) = getData(uint256(parentNode));
+        if (maxExpiry < block.timestamp) {
```
Arachnid, line 951: You can probably skip this too, though I'm less confident in saying that...

Arachnid (changes requested):

PR #158, opened 2022-11-21
Some edits to NameWrapper.sol and NameWrapper.js

nxt3d:
- Name is only used for an event, not stored, so I think we should change the name. 
- Remove sub.fuses.eth from a test. It is out of place.


jefflau:
Closing as stale

PR #159, opened 2022-12-01
Unwrapped burning protection

jefflau:
* Add back in `isWrapped()` for `setSubnodeOwner/Record`
* Add protection against setting child fuses for unwrapped names in `setChildFuses()`
* Fix tests
* Add test to ensure TLDs can still be recovered if they expire via root calling ens.setSubnodeOwner()

Fixes:
https://github.com/code-423n4/2022-11-ens-findings/issues/14
https://github.com/code-423n4/2022-11-ens-findings/issues/16

These fixes were bundled together as they were written before Zhuo was officially added to the audit. They are also interrelated on the idea that: Names that are implicitly unwrapped need to be explicitly handled by the NameWrapper.

isWrapped() should understand when a subdomain has been unwrapped and needs to be re-wrapped. Even if it has been unwrapped by the parent using the registry.

For .eth names that have been implicitly unwrapped, if they are put into that state, they are no longer able to affect subdomains by calling `setChildFuses()`, because `setChildFuses()` is now protected against being called by names that do not have the NameWrapper as the owner on the .eth registrar.

### DIFF
--- a/contracts/wrapper/NameWrapper.sol
+++ b/contracts/wrapper/NameWrapper.sol
@@ -22,6 +22,7 @@ error LabelTooLong(string label);
 error IncorrectTargetOwner(address owner);
 error CannotUpgrade();
 error OperationProhibited(bytes32 node);
+error NameIsNotWrapped();
 
 contract NameWrapper is
     Ownable,
@@ -129,7 +130,9 @@ contract NameWrapper is
 
         bytes32 labelhash = _getEthLabelhash(bytes32(id), fuses);
         if (labelhash != bytes32(0)) {
-            expiry = uint64(registrar.nameExpires(uint256(labelhash))) + GRACE_PERIOD;
+            expiry =
+                uint64(registrar.nameExpires(uint256(labelhash))) +
+                GRACE_PERIOD;
         }
 
         if (expiry < block.timestamp) {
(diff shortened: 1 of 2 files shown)


contracts/wrapper/NameWrapper.sol
```diff
@@ -527,7 +538,7 @@ contract NameWrapper is
         bytes memory name = _saveLabel(parentNode, node, label);
         expiry = _checkParentFusesAndExpiry(parentNode, node, fuses, expiry);
 
-        if (ownerOf(uint256(node)) == address(0)) {
```
Arachnid, line 530: Maybe we could incorporate the `isWrapped` check into `onlyTokenOwner`? That way we avoid duplicate operations and get it everywhere we need.

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
nxt3d: We already check ens.owner(node) in canCallSetSubnodeOwner, so I don't thin this is necessary?

ZhangZhuoSJTU: Imho, the check here (i.e., `isWrapped`) is slightly different from the one in `canCallSetSubnodeOwner`. Specifically

+ The [ens owner check](https://github.com/ensdomains/ens-contracts/blob/19582673577cc7b97b6db4223e0f523f04bda514/contracts/wrapper/NameWrapper.sol#L683) in `canCallSetSubnodeOwner` guarantees that a `parentNode` cannot create subnodes if its `CANNOT_CREATE_SUBDOMAIN` is burnt

+ The check here is to guarantee that, for a given `node`, every time it has a fuse to burn, the `ens.owner(node)` should be `address(this)`. It is to prevent some potential hacks.

I am not really an ENS professional, and hence may be wrong here. Please kindly correct me if so.

Arachnid (approved):
Okay, this LGTM. I only wonder if we should be checking this in more places. Should it be possible to transfer an unwrapped name, for instance?

PR #162, opened 2022-12-06
Prevent unwrapped names via re-registering previously wrapped names

jefflau:
* Fixes issue where registering a previously wrapped name allows ownership on it in the NameWrapper

### DIFF
--- a/contracts/wrapper/NameWrapper.sol
+++ b/contracts/wrapper/NameWrapper.sol
@@ -129,7 +129,19 @@ contract NameWrapper is
 
         bytes32 labelhash = _getEthLabelhash(bytes32(id), fuses);
         if (labelhash != bytes32(0)) {
-            expiry = uint64(registrar.nameExpires(uint256(labelhash))) + GRACE_PERIOD;
+            expiry =
+                uint64(registrar.nameExpires(uint256(labelhash))) +
+                GRACE_PERIOD;
+            // if owner in registrar is not the wrapper, zero out the owner
+            try registrar.ownerOf(uint256(labelhash)) returns (
+                address registrarOwner
+            ) {
+                if (registrarOwner != address(this)) {
+                    owner = address(0);
+                }
+            } catch {
+                // do nothing if name is unregistered
+            }
         }
 
         if (expiry < block.timestamp) {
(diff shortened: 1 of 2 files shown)


contracts/wrapper/NameWrapper.sol
```diff
@@ -129,7 +129,19 @@ contract NameWrapper is
 
         bytes32 labelhash = _getEthLabelhash(bytes32(id), fuses);
         if (labelhash != bytes32(0)) {
-            expiry = uint64(registrar.nameExpires(uint256(labelhash))) + GRACE_PERIOD;
+            expiry =
+                uint64(registrar.nameExpires(uint256(labelhash))) +
+                GRACE_PERIOD;
+            // if owner in registrar is not the wrapper, zero out the owner
+            try registrar.ownerOf(uint256(labelhash)) returns (
```
Arachnid, line 136: You can probably avoid the need for the try/catch if you just check if expiry + grace period is in the future.

contracts/wrapper/NameWrapper.sol
```diff
@@ -127,9 +127,19 @@ contract NameWrapper is
...
-        if (labelhash != bytes32(0)) {
-            expiry = uint64(registrar.nameExpires(uint256(labelhash))) + GRACE_PERIOD;
+        bytes32 labelHash = _getEthLabelhash(bytes32(id), fuses);
+        if (labelHash != bytes32(0)) {
+            uint64 registrarExpiry = uint64(
+                registrar.nameExpires(uint256(labelHash))
+            );
+            expiry = registrarExpiry + GRACE_PERIOD;
+            // if owner in registrar is not the wrapper, zero out the owner
+            if (
```
Arachnid, line 137: Hm, does this mean that a name that's subject to the vulnerability will briefly appear to have an owner again during the grace period? We should probably zero out the owner if the expiry is in the past even if it *is* in grace.

jefflau: Isn't this already doing that?

Arachnid: My mistake.

Arachnid (approved):

PR #164, opened 2022-12-08
Mitigations for subnode creation

jefflau:
1. Protects subdomains from being burnt whilst unexpired and then recreated (if CANNOT_CREATE_SUBDOMAIN) is already burnt
2. Protects subdomains from unexpiring and then being created again by a parent that has CANNOT_CREATE_SUBDOMAIN burnt
3. Retain the ability for subdomains to be unwrapped and retain their protection with PCC

Fixes:
https://github.com/code-423n4/2022-11-ens-findings/issues/8
https://github.com/code-423n4/2022-11-ens-findings/issues/7

Fixes merged as one PR as it affects whether or not a subdomain is considered "created" or not.

### DIFF
--- a/contracts/wrapper/NameWrapper.sol
+++ b/contracts/wrapper/NameWrapper.sol
@@ -129,7 +129,9 @@ contract NameWrapper is
 
         bytes32 labelhash = _getEthLabelhash(bytes32(id), fuses);
         if (labelhash != bytes32(0)) {
-            expiry = uint64(registrar.nameExpires(uint256(labelhash))) + GRACE_PERIOD;
+            expiry =
+                uint64(registrar.nameExpires(uint256(labelhash))) +
+                GRACE_PERIOD;
         }
 
         if (expiry < block.timestamp) {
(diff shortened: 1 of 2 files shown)


contracts/wrapper/NameWrapper.sol
```diff
@@ -672,27 +671,40 @@ contract NameWrapper is
...
+        // check if the registry owner is 0 and expired
+        // check if the wrapper owner is 0 and expired
+        // If either, then check parent fuses for CANNOT_CREATE_SUBDOMAIN
+        bool expired = subnodeExpiry < block.timestamp;
+        if (
+            expired &&
+            // protects a name that has been burnt and doesn't allow the parent to take control by recreating it if unexpired
+            (ens.owner(subnode) == address(0) ||
+                // protects a name that has been unwrapped with PCC and doesn't allow the parent to take control by recreating it if unexpired
+                subnodeOwner == address(0))
```
Arachnid, line 697: If you swap these two checks, it will save gas in the instance that `subnodeOwner` is 0 thanks to short-circuiting.

Arachnid (changes requested):

Arachnid (approved):

PR #173, opened 2022-12-21
New parent-controlled fuse to allow names to extend their own expiry

serenae-fansubs:
Just an idea I had. Feel free to close if this doesn't make sense or introduces other issues.

Reserves a new parent-controlled fuse that allows owners of subnames to extend their own expiry.

This gives a first-class easy way to allow perpetual "unruggable" renewals for subname owners, without needing to use some other contract to indirectly call setChildFuses.

This will not be burned on .eth 2LDs, and it doesn't matter there anyway since the expiry will always inherit from the registrar in that case.


nxt3d:
I have also been researching ways to solve this issue and have come up with some potential solutions. One idea I came up with is to use an external controller called a "lightcontroller." The lightcontroller only has the authority to extend the name and nothing else. The parent name owner can assign the lightcontroller to a subname, which is then guaranteed to be able to be renewed indefinitely according to the conditions set by the lightcontroller. This system allows for flexible management of subname renewals through smart contracts. Lightcontrollers can have administrator functionality or not. For example, a marketplace could create its own lightcontroller to manage renewal prices and royalties for individual names. Alternatively, it could be possible to implement different prices for different length names or have other complex pricing schemes.

https://github.com/nxt3d/ens-contracts/blob/NameWrapper-LightControllers/contracts/wrapper/LightcontrollerService.sol

serenae-fansubs:
I think that's another great potential solution! I feel that anything that involves needing to approve a separate contract is going to be too confusing and too much work for the average person.

This PR however, is pretty dead-simple, and builds it directly into the NameWrapper. This may not have made sense in previous iterations, but now that we have the concept of **parent-controlled fuses**, I think this would be a perfect example of a "perk" that a parent can allow for a child name.

And all they need to do is burn this one fuse on the child!

I think your lightcontroller idea is useful in general for other things though. Of course, if someone is spinning up their own subdomain registrar, all that stuff you describe is going to be built into the registrar contract itself. No need for a separate controller, unless you want to separate out knobs to tweak pricing/etc the way the .eth registrar does.

jefflau:
This is interesting. It creates an easy way to make a name a 'forever' name. The parent name would need to expire to unburn this fuse so you could almost guarantee you'll always have control over the name. We could make it so it isn't restricted to the owner of the name, but the parent + owner w/ fuse burnt can both call this function.

serenae-fansubs:
Good point!

I updated the PR so that the parent owner can call `setExpiry` too.

Arachnid:
This looks like a good idea to me.

jefflau:
@serenae-fansubs As a next step can you write some test cases for this PR?

contracts/wrapper/NameWrapper.sol
```diff
@@ -397,6 +397,53 @@ contract NameWrapper is
...
+            }
+        }
+
+        // max expiry is set to the expiry of the parent
+        (, , uint64 maxExpiry) = getData(
+            uint256(parentNode)
+        );
+        expiry = _normaliseExpiry(expiry, oldExpiry, maxExpiry);
+
+        _setFuses(node, owner, fuses, expiry);
```
jefflau, line 443: I think we can call `_setData` directly as we won't be burning any fuses

jefflau (changes requested):

serenae-fansubs:
@jefflau Done!

jefflau:
I think this is ready.  I think it may make sense to split the events now to event `ExpiryExtended()` and FusesSet()`, and call both events in `setChildFuses()`. What do you think @Arachnid?

jefflau:
I've merged this PR into a local branch so I can make the above event changes. Please refer to it here:

https://github.com/ensdomains/ens-contracts/pull/184

PR #177, opened 2022-12-28
Add token approvals to PublicResolver.sol

nxt3d:
The current version of the Public Resolver only allows owners to approve other addresses to update all of their records on their behalf. However, this proposed change (referred to as "token approvals") would allow an owner to approve a delegate to make changes to the records of a specific name. This would be useful for companies that want to retain control over their subnames but still allow employees to make updates to individual subnames assigned to them. For example, an employee could update all of the records for a subname like "greg.technet.eth," but the company would retain ownership and be able to revoke the approved wallet address at any time.

Another potential use case for token approvals is if a parent name is wrapped in a contract. In this scenario, the contract could approve the previous owner of the name in the Public Resolver, allowing them to continue using the name and making updates through the ENS app.

### DIFF
--- a/contracts/resolvers/PublicResolver.sol
+++ b/contracts/resolvers/PublicResolver.sol
@@ -139,7 +139,8 @@ contract PublicResolver is
         if (owner == address(nameWrapper)) {
             owner = nameWrapper.ownerOf(uint256(node));
         }
-        return owner == msg.sender || isApprovedForAll(owner, msg.sender);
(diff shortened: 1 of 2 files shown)


Arachnid (commented):
This seems reasonable to me. @jefflau your thoughts?

jefflau:
This looks good to me too.

jefflau (approved):

PR #180, opened 2023-01-03
Feature/added public resolver uri fallback

TateB:


contracts/resolvers/PublicResolverWithFallback.sol
```diff
@@ -0,0 +1,58 @@
...
+     * Returns the text data associated with an ENS node and key.
+     * @param node The ENS node to query.
+     * @param key The text data key to query.
+     * @return record The associated text data.
+     */
+    function text(
+        bytes32 node,
+        string calldata key
+    ) external view virtual override returns (string memory record) {
+        record = versionable_texts[recordVersions[node]][node][key];
```
Arachnid, line 47: Can you call super here rather than fetching it directly?

contracts/resolvers/PublicResolverWithFallback.sol
```diff
@@ -0,0 +1,58 @@
...
+        if (
+            bytes(record).length == 0 &&
+            bytes(fallback_text_uris[key]).length > 0
+        ) {
+            record = string.concat(
+                fallback_text_uris[key],
+                string(abi.encodePacked(node))
+            );
```
Arachnid, line 55: This will result in concatenating text and binary data together. Presumably that's not what you intended?

Arachnid (changes requested):

Arachnid (approved):

PR #181, opened 2023-01-03
Store eth expiry

jefflau:
1. Remove registrar.nameExpires() from everything and check expiry from getData() just like a normal wrapped name

1. renew() and wrapETH2LD() update expiry based onregistrar.nameExpires()`

1. renew()  if name is not wrapped** does not update eth expiry but just passes the call directly to registrar.renew and then returns without changing the state within the wrapper.

The idea is to not automatically update the expiry inside the wrapper by calling the registrar, but instead only updating it on wrapETH2LD() and renew(). This means if anyone calls the old controller, it will not extend expiry and allow them to use the name within wrapping.

**Not wrapped is defined as: `registrar.ownerOf()` OR `registry.owner()` are not the Name Wrapper contract OR `NameWrapper.ownerOf()` is zero`

### DIFF
--- a/contracts/wrapper/ERC1155Fuse.sol
+++ b/contracts/wrapper/ERC1155Fuse.sol
@@ -228,7 +228,17 @@ abstract contract ERC1155Fuse is ERC165, IERC1155, IERC1155MetadataURI {
             (uint256(expiry) << 192);
     }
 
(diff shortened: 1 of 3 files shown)


Arachnid (commented):
Looks good, but needs a test that fails pre-change.

PR #183, opened 2023-01-10
Used E165Storage instead an inheritance

dkulyk:
Calling "supportsInterface" for realization with mapping costs less than inherited calls with all those profiles.


Arachnid:
Do you have any numbers to back this claim up?

PR #184, opened 2023-01-11
Extend expiry fuse

jefflau:


serenae-fansubs:
@jefflau The `setExpiry()` method doesn't alter fuses, so I'm not sure if it matters, but should this still revert with `NameIsNotWrapped` anyway if the name is unregistered/unwrapped? I made some updates for this in case you think it's needed here.

FYI I tried at first to create a branch off of this one and commit to that, but I don't think I have the proper permissions to push commits to this repository.

So instead I had to create a separate PR from my fork: #185

jefflau:
@serenae-fansubs I've split the events up and changed the name to extend instead. Can you take a look?

serenae-fansubs:
@jefflau Looks good to me! The only thing that needs changing is the wrapper readme. Wherever it says `setExpiry`, switch that out with `extendExpiry`.

contracts/wrapper/NameWrapper.sol
```diff
@@ -413,10 +413,56 @@ contract NameWrapper is
...
+        if (!canModifyParentName && !canModifyName(node, msg.sender)) {
+            revert Unauthorised(node, msg.sender);
+        }
+
+        (address owner, uint32 fuses, uint64 oldExpiry) = getData(
+            uint256(node)
+        );
+        if (owner == address(0) || ens.owner(node) != address(this)) {
+            revert NameIsNotWrapped();
+        }
```
Arachnid, line 447: Shouldn't we be doing this check elsewhere (or not at all)? We don't want to be doing it on every method separately.

contracts/wrapper/NameWrapper.sol
```diff
@@ -413,10 +413,56 @@ contract NameWrapper is
...
+
+        // the parent owner can always set expiry, so no need to check fuses
+        if (!canModifyParentName) {
+            // revert if CAN_EXTEND_EXPIRY parent-controlled fuse has not been burned
+            if (fuses & CAN_EXTEND_EXPIRY == 0) {
+                revert OperationProhibited(node);
+            }
+        }
```
Arachnid, line 455:
```suggestion
        // Either CAN_EXTEND_EXPIRY must be set, or the caller must have permission to modify the parent name
        if (!canModifyParentName && fuses & CAN_EXTEND_EXPIRY == 0) {
            revert OperationProhibited(node);
        }
```

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
serenae-fansubs: I believe it is being done in several other places too, this logic was just added recently for the audit mitigations: [https://github.com/ensdomains/ens-contracts/pull/170/files#diff-8f057ac6433f8d6754bd5e1e9852faaf2d182b21cad6f5869f9041169c729c77R508](https://github.com/ensdomains/ens-contracts/pull/170/files#diff-8f057ac6433f8d6754bd5e1e9852faaf2d182b21cad6f5869f9041169c729c77R508)

Arachnid (approved):

PR #185, opened 2023-01-11
Preventing setExpiry from being called on unregistered/unwrapped names.

serenae-fansubs:
Since the `setExpiry()` method does not alter fuses, perhaps this isn't needed. But I went ahead and added the `NameIsNotWrapped` changes that you made elsewhere for audit mitigations.


jefflau (approved):

jefflau:
I think this makes sense and i've merged it into my branch.

PR #195, opened 2023-01-27
Fix setFuses return data

nxt3d:
This PR addresses an issue with the setFuses function where previously set fuses were not being included in the returned result. A test has been added to ensure the returned fuses are accurate.

### DIFF
--- a/contracts/wrapper/NameWrapper.sol
+++ b/contracts/wrapper/NameWrapper.sol
@@ -400,7 +400,7 @@ contract NameWrapper is
      * @notice Sets fuses of a name
      * @param node Namehash of the name
      * @param ownerControlledFuses Owner-controlled fuses to burn
-     * @return New fuses
+     * @return Old fuses
      */
 
     function setFuses(bytes32 node, uint16 ownerControlledFuses)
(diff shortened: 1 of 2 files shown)


Arachnid:
I wonder if it makes more sense to leave this as-is and update the docs. The caller can always calculate the new fuses from the old ones, but they can't do the inverse, so this gives them more information.

jefflau:
> I wonder if it makes more sense to leave this as-is and update the docs. The caller can always calculate the new fuses from the old ones, but they can't do the inverse, so this gives them more information.

I think this makes sense. @nxt3d would you mind updating the docs?

nxt3d:
> I wonder if it makes more sense to leave this as-is and update the docs. The caller can always calculate the new fuses from the old ones, but they can't do the inverse, so this gives them more information.

The function just returns the parameter, ownerControlledFuses, it currently serves no purpose.

jefflau:
I think Nick meant to return oldFuses, not the parameter.

nxt3d:
> I think Nick meant to return oldFuses, not the parameter.

Ok, got it, I will update the PR as well as the docs.

nxt3d:
I updated this PR to now return old fuses instead of the new fuses. @jefflau

PR #196, opened 2023-01-27
Extend expiry edits

nxt3d:
This PR updates the extendExpiry function by replacing the use of _setData with super._setData, as the latter is sufficient for extending expiry without the need to check if fuses can be burned. Additionally, by using an if/else block, it avoids unnecessary calls to canModifyName when a parent is setting the expiry, resulting in a savings of around 200 gas.


jefflau (changes requested):
Bypassing the fuses check by calling super._setData makes me nervous. I’m not sure if it’s worth it as we’re not saving much gas. Best to be defensive in my opinion.

nxt3d:
I removed "super." from super._setData to avoid a call to super. @jefflau

jefflau:
We are not reverting with the correct custom errors if your if else statement.

I think it would be best to rewrite it as follows:

```
function extendExpiry(
    bytes32 parentNode,
    bytes32 labelhash,
    uint64 expiry
) public returns (uint64) {

    // this flag is used later, when checking fuses
    bool canModifyParentName = canModifyName(parentNode, msg.sender);
    // only allow the owner of the name or owner of the parent name
    if(canModifyParentName){
        _setData(node, owner, fuses, expiry);
        emit ExpiryExtended(node, expiry);
    } 

    if (!canModifyName(node, msg.sender)) {
    }

    (address owner, uint32 fuses, uint64 oldExpiry) = getData(
        uint256(node)
    );

    // Either CAN_EXTEND_EXPIRY must be set, or the caller must have permission to modify the parent name
    if (fuses & CAN_EXTEND_EXPIRY == 0) {
    }

    // max expiry is set to the expiry of the parent
    (, , uint64 maxExpiry) = getData(uint256(parentNode));
    expiry = _normaliseExpiry(expiry, oldExpiry, maxExpiry);

    _setData(node, owner, fuses, expiry);
    emit ExpiryExtended(node, expiry);
    return expiry;
}
```

jefflau:
We need a revert for when the fuse isn't burn, but the owner can modify the name. I think it's probably better written as follows:

```solidity
function extendExpiry(
    bytes32 parentNode,
    bytes32 labelhash,
    uint64 expiry
) public returns (uint64) {

    (address owner, uint32 fuses, uint64 oldExpiry) = getData(
        uint256(node)
    );
    // max expiry is set to the expiry of the parent
    (, , uint64 maxExpiry) = getData(uint256(parentNode));
    expiry = _normaliseExpiry(expiry, oldExpiry, maxExpiry);

    // this flag is used later, when checking fuses
    bool canModifyParentName = canModifyName(parentNode, msg.sender);
    // only allow the owner of the name or owner of the parent name
    if(canModifyParentName){
        _setData(node, owner, fuses, expiry);
        emit ExpiryExtended(node, expiry);
        return expiry;
    } 

    if (!canModifyName(node, msg.sender)) {
    }

    // Either CAN_EXTEND_EXPIRY must be set, or the caller must have permission to modify the parent name
    if (fuses & CAN_EXTEND_EXPIRY == 0) {
    }

    // If can modify name and fuse is burnt
    _setData(node, owner, fuses, expiry);
    emit ExpiryExtended(node, expiry);
    return expiry;
}
```

nxt3d:
I incorporated the changes, but I was still not satisfied with the gas usage, so I got it down to these numbers by adding an internal function canModifyNameWithData. The function is very easy to audit, because it is just a copy and paste of canModifyName with the getData call taken out, which is often redundant. 

Before>  · 33021  ·  50311  ·  35565  

Now> · 32275  ·  49906  ·  34999

contracts/wrapper/NameWrapper.sol
```diff
@@ -212,6 +212,28 @@ contract NameWrapper is
...
+    /**
+     * @notice Check if the address is the owner or approved by owner.
+     * @param owner The owner of the name.
+     * @param fuses The fuses of the name.
+     * @param expiry The expiry of the name.
+     * @param addr The address address to check permissions for.
+     * @return whether or not is owner or approved.
+     */
+
+    function canModifyNameWithData(address owner, uint32 fuses, uint64 expiry, address addr)
```
jefflau, line 225:
```suggestion
    function _canModifyNameWithData(address owner, uint32 fuses, uint64 expiry, address addr)
```

jefflau: Can you use this internal function inside `canModifyName`?

jefflau (changes requested):

contracts/wrapper/NameWrapper.sol
```diff
@@ -432,29 +454,37 @@ contract NameWrapper is
...
+
+        // Allow the owner of the parent name to extend the expiry.
+        if(canModifyNameWithData(parentOwner, parentFuses, parentExpiry, msg.sender)){
+            _setData(node, owner, fuses, expiry);
+            emit ExpiryExtended(node, expiry);
+            return expiry;
+        } 
+
+        // If the caller is not the parent then check to make sure the caller is the owner of the name.
+        if (!canModifyNameWithData(owner, fuses, oldExpiry, msg.sender)) {
```
jefflau, line 474: Can you make a note in the comments that this needs to be oldExpiry. If we accidentally put expiry here in a future update, we might accidentally let names renew when they are already expired.

contracts/wrapper/NameWrapper.sol
nxt3d: Ok, but it will just cost more gas.

contracts/wrapper/NameWrapper.sol
nxt3d: Ok

nxt3d: hmm, not sure I fully get this one. Can you write the comment and then I can just insert it? @jefflau

contracts/wrapper/NameWrapper.sol
```diff
@@ -432,29 +454,37 @@ contract NameWrapper is
...
+        expiry = _normaliseExpiry(expiry, oldExpiry, parentExpiry);
+
+        // Allow the owner of the parent name to extend the expiry.
+        if(canModifyNameWithData(parentOwner, parentFuses, parentExpiry, msg.sender)){
+            _setData(node, owner, fuses, expiry);
+            emit ExpiryExtended(node, expiry);
+            return expiry;
+        } 
+
+        // If the caller is not the parent then check to make sure the caller is the owner of the name.
```
jefflau, line 473:
```suggestion
        // If the caller is not the parent then check to make sure the caller is the owner of the name.
        // _canModifyNameWithData explicitly takes oldExpiry, not the new expiry passed to this function
```

nxt3d:
@jefflau Ok, I committed your comment change.

contracts/wrapper/NameWrapper.sol
jefflau: > @jefflau Ok, I committed your comment change.

Okay, just need the internal function changes resolved and prefixed with an underscore. And move it down to the internal functions section of the contract

nxt3d: ok

nxt3d:
I rebased using the changes from PR #217.

jefflau:
NameWrapper has been deployed to mainnet, so closing this PR

PR #197, opened 2023-01-30
token delegation from multi-delegate to single

nxt3d:
This PR changes the token delegation in the public resolver from multi-delegation to single delegation. The advantage with single delegation is that it is much easier to discover and clear old delegations.

### DIFF
--- a/contracts/resolvers/PublicResolver.sol
+++ b/contracts/resolvers/PublicResolver.sol
@@ -40,22 +40,23 @@ contract PublicResolver is
      * A mapping of operators. An address that is authorised for an address
      * may make any changes to the name that the owner could, but may not update
      * the set of authorisations.
-     * owner => operator
+     * (owner, operator) => approved
      */
-    mapping(address => address) private _operatorApprovals;
+    mapping(address => mapping(address => bool)) private _operatorApprovals;
 
     /**
      * A mapping of delegates. The delegate that is set by an owner
      * for a name may make changes to the name's resolver, but may not update
      * the set of token approvals.
-     * (owner, name) => delegate
+     * (owner, name, delegate) => approved
      */
     mapping(address => mapping(bytes32 => address)) private _tokenApprovals;
 
     // Logged when an operator is added or removed.
     event ApprovalForAll(
         address indexed owner,
-        address indexed operator
+        address indexed operator,
+        bool approved
     );
 
     // Logged when a delegate is approved or  an approval is revoked.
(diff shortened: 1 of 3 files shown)


nxt3d:
INameWrapper was missing a lot of public functions, including isWrapped. I added all the public functions (except for onERC721Received) of NameWrapper into INameWrapper, so that anyone can use it to call functions on the NameWrapper.

Arachnid:
This is a breaking change, so I'd need fairly convincing evidence that the current behaviour isn't being used before approving it.

branzinho-nt:
I have been following the latest changes regarding the NameWrapper, since our team at thename.space is building on top of it, and have also looked into the PublicResolver changes. This latest one is fine with us, and will have no impact on our team

stevegachau:
We weren't using multi-delegation so this change will not impact our work.

yayashuxue:
Hey, this is julieshi.eth, I work on https://twitter.com/org3_eth and https://doc3.me. This change does not affect my projects. Here are my thoughts on this change:
1. Echo with premm.eth there is no way to iterate over a mapping structure, so it would be hard to track who are still the delegates.
2. I think a multi-sig wallet that needs single approval in a single delegation scenario is equivalent to multi-delegation. So there is a way to work around it even after the change.
3. My question would be: Is there a way that to keep multi-delegation while it can also be tracked (#1 problem solved)?

nxt3d:
> 3\. My question would be: Is there a way to keep multi-delegation while it can also be tracked (first problem solved)?

You can use an Enumerable Set:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol

https://delegate.cash uses this method. This adds a lot of complexity however, and extra gas.

hodlthedoor:
Interesting. 

I've not really done anything specific with the public resolvers so this change wouldn't affect me

0xFlicker:
I'm fine with a single delegate. It matches what you see elsewhere on ENS

Arachnid:
It strikes me as weird that this would leave us with two forms of approval, where `setApprovalForAll` allows any number of approved users, and `approve` only allows one per name. We need to harmonise this (and the names and events) if we're to make a change.

Arachnid:
The current state isn't ideal either, since we can't remove approvals.

nxt3d:
As we are not bound by the ERC1155 standard, we can restrict operator approvals to a single operator. This simplifies the process of replacing an old operator and eliminates the need for events to track approvals, as clearing out an old operator and setting a new one can be done easily in one transaction.

yayashuxue:
Update: My latest development actually is involved with setApprovalForAll. However, the change should be ok for now as the user only needs to have one delegate at a time. However, I am wondering if in the future a domain owner needs different delegates to satiate different needs, user adding a new delegate may lead to an unexpected outcome, removing the right for another delegate that the user has been using for a while.

jefflau:
So I've been mulling this over the past week. My thoughts are as follows:

- We use ERC-1155 style approvals for all other ENS contracts:  NameWrapper, Registry, Registrar.
- Resolvers are not ERC-1155 tokens but all other contracts use this approval mechanism and it's a standard to use multi-approvals. We would be diverging from the industry standard AND the current ENS standard.
- We shouldn't use multi AND single at the same time as it's confusing, so we shouldn't make this single and then leave multi-delegation for `setApprovalForAll`. But moving to single delegation for `setApprovalForAll` means we are breaking from convention
- If we break from convention we should have a good reason for doing so. Single delegation means we can remove and add an approval in one transaction. It also means we can easily check who is delegated to on-chain without knowing their address. However we also lose simple multi delegation. Currently with NFTs we are use to giving approvals to many different services. If delegated records become a staple in lots of decentralised applications, single delegation feels quite limiting. Most of the usecases for getting a list to check who is approved is done off-chain (e.g. in a UI to show who is approved). For off-chain usecases we can just simply use the graph or some other event tracking service.

All in all, I'm leaning towards not merging this as I can't see strong argument to break from standards.

nxt3d:
I can take out single approvals for setApprovalForAll, but token approvals are a ERC721 standard and they are single.

nxt3d:
I reverted the changes to setApprovalForAll. Now both functions adhere to their standards, ERC1155 and ERC721.

PR #198, opened 2023-02-03
Remove SafeMath

nxt3d:
@jefflau //  None of the functions, including mul, div, sub, add, and mod, are used in this contract.


jefflau:
Thanks!

PR #203, opened 2023-02-09
Implement offchain DNS Registrar support

Arachnid:
This PR implements a new resolver that can be set as a wildcard resolver for any TLD other than .eth. Once set, it resolves all subdomains that do not exist in the registry by making CCIP-Read calls to a DNS gateway, which returns signed DNS TXT records. It verifies those using the DNSSEC oracle in a call operation, and thus makes it possible to find the resolver for a name without any gas required to set it.

TXT records are expected to be on the name in question, and take the format:
```
ENS1 <name-or-address> [context]
```

`name-or-address` is either an ENS name or the 0x-prefixed address of the resolver to use to resolve this DNS name. If an ENS name is specified, it must be in the ENS registry - eg, it cannot be resolved using wildcard resolution or via CCIP-Read.

The resolver will be queried using ERC165. If it supports the new `IExtendedDNSResolver` interface, `resolve()` is called on it, passing the namehash, encoded query, and `context`. If it supports `IExtendedResolver`, `resolve()` is called with namehash and encoded query, but any context data is ignored. If neither is supported, the resolver is called with the encoded query directly.

This makes it possible to use either a regular resolver (for example, the `OwnedResolver`; `PublicResolver` will not work here, as it cannot check ownership of offchain names) with records stored onchain, or a DNS-specific resolver that understands how to parse `context` and use it to resolve the query without any onchain write operations at all. A standardised implementation of this will be forthcoming shortly.


contracts/resolvers/profiles/ExtendedResolver.sol
```diff
@@ -0,0 +1,19 @@
...
+        external
+        view
+        returns (bytes memory)
+    {
+        (bool success, bytes memory result) = address(this).staticcall(data);
+        if(success) {
+            return result;
+        } else {
+            assembly {
+                revert(add(result, 0x20), mload(result))
```
makoto, line 15: can you add comment on what's doing?

test/dnsregistrar/TestOffchainDNSResolver.js
```diff
@@ -0,0 +1,206 @@
...
+    expect(ethers.utils.defaultAbiCoder.decode(['address'], result)[0]).to.equal(testAddress);
+  })
+
+  it('handles calls to resolveCallback() with extra data and a legacy resolver', async function() {
+    const name = "test.test";
+    const testAddress = '0xfefeFEFeFEFEFEFEFeFefefefefeFEfEfefefEfe';
+    await ownedResolver.setAddr(namehash.hash(name), testAddress);
+    const pr = await PublicResolver.at(offchainResolver.address);
+    const callData = pr.contract.methods['addr(bytes32)'](namehash.hash(name)).encodeABI();
+    const result = await doResolveCallback(name, [`ENS1 ${ownedResolver.address} blah`], callData);
```
makoto, line 141: what does the `blah` suppose to do? Is this just testing that the callback returns the same result with or without the `blah` text without throwing an error?

test/dnsregistrar/TestOffchainDNSResolver.js
```diff
@@ -0,0 +1,206 @@
...
+  it('handles calls to resolveCallback() with valid DNS TXT records containing a name', async function() {
+    // Configure dnsresolver.eth to resolve to the ownedResolver so we can use it in the test
+    await root.setSubnodeOwner(ethers.utils.id('eth'), accounts[0]);
+    await ens.setSubnodeOwner(namehash.hash('eth'), ethers.utils.id('dnsresolver'), accounts[0]);
+    await ens.setResolver(namehash.hash('dnsresolver.eth'), ownedResolver.address);
+    await ownedResolver.setAddr(namehash.hash('dnsresolver.eth'), ownedResolver.address);
+
+    const name = "test.test";
+    const testAddress = '0xfefeFEFeFEFEFEFEFeFefefefefeFEfEfefefEfe';
+    await ownedResolver.setAddr(namehash.hash('test.test'), testAddress);
```
makoto, line 154: can you just put `await ownedResolver.setAddr(namehash.hash(name), testAddress);` for the consistency unless this line is doing something different to other test?

contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -0,0 +1,163 @@
+
```
makoto, line 1: Do you want to add the below?

```
pragma solidity ^0.8.4;
```

makoto:
Can you add a deploy script or deploy to tenderly fork so we can test on the real data?

makoto:
Is `context` where you set an actual record like `addr`, `contanthash`,  `text`, `abi`, `pubkey`, etc?

test/dnsregistrar/TestOffchainDNSResolver.js
Arachnid: That's right.

Arachnid:
>  Can you add a deploy script or deploy to tenderly fork so we can test on the real data?

I'll add a deploy script.

> Is context where you set an actual record like addr, contanthash, text, abi, pubkey, etc?

The interpretation of `context` is up to the resolver contract, but that's the most obvious use for it, yes.

> Assuming we have more PR to complete the whole featureset, can you create an integration branch and make PR against the integration branch, rather than merging directly into master?

The code in this PR is useful on its own, and there are an unlimited number of possible resolver implementations that work with it. I don't think there's any point in holding it back.

Arachnid:
@makoto PTAL.

makoto (approved):

PR #206, opened 2023-02-11
Change BaseRegistrar and PriceOracle exports to IBaseRegistrar and IPriceOracle

0xcharchar:
Was upgrading from 0.0.8 to 0.0.17 and found this reference was broken. Based on dates, it may have been broken since 0.0.12 but I did not confirm.

I saw that there was another PR (#103), which is likely a better long-term solution, but wanted to get a quick fix up first.


Arachnid (approved):
Thanks!

PR #207, opened 2023-02-14
Fix NameWrapper upgrade

TateB:


contracts/wrapper/NameWrapper.sol
```diff
@@ -655,6 +702,30 @@ contract NameWrapper is
         }
     }
 
+    function setSubnodeRecordFromUpgrade(
```
jefflau, line 705: Probably don't need this func

jefflau (approved):

PR #208, opened 2023-02-18
Fix Name Wrapper interface issues

nxt3d:
This PR fixes a number of issues with the INameWrapper interface including:
- Removing a duplicate INameWrapper interface
- Adding missing functions to the INameWrapper, e.g. isWrapped
- Removing unnecessary interface override statements. 

In order to get the tests to pass I needed to clear old build data. I used:
```
rm -fr node_modules build cache artifacts
yarn
```
@jefflau

### DIFF
--- a/contracts/resolvers/PublicResolver.sol
+++ b/contracts/resolvers/PublicResolver.sol
@@ -11,10 +11,7 @@ import "./profiles/NameResolver.sol";
 import "./profiles/PubkeyResolver.sol";
 import "./profiles/TextResolver.sol";
 import "./Multicallable.sol";
-
-interface INameWrapper {
-    function ownerOf(uint256 id) external view returns (address);
-}
+import "../wrapper/INameWrapper.sol";
 
 /**
  * A simple resolver anyone can use; only allows the owner of a node to set its
--- a/contracts/wrapper/INameWrapper.sol
+++ b/contracts/wrapper/INameWrapper.sol
@@ -5,6 +5,7 @@ import "../registry/ENS.sol";
 import "../ethregistrar/IBaseRegistrar.sol";
 import "@openzeppelin/contracts/token/ERC1155/IERC1155.sol";
 import "./IMetadataService.sol";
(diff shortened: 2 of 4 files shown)


jefflau:
I've been thinking about `isWrapped()`, and I think we should probably make at least this version of it internal, since it does not check for .eth names.

Also note upgrade contract has been changed (and just merged) due to an issue we found, so this would change some of those function signatures.

> Removing unnecessary interface override statements.

I thought override/virtual was required on functions that are set in the interface and then overwritten?

nxt3d:
>I've been thinking about isWrapped(), and I think we should probably make at least this version of it internal, since it does not check for .eth names. Possibly we could expose an offchain version that does all the checks and does not have to be gas efficient

ok

> Also note upgrade contract has been changed (and just merged) due to an issue we found, so this would change some of those function signatures.

noted

> I thought override/virtual was required on functions that are set in the interface and then overwritten?

It's not necessary since 0.8.8
https://docs.soliditylang.org/en/v0.8.18/contracts.html#function-overriding

>Also, can we separate out the Resolver.sol interfaces, so we can look at them separately instead of bundled with NameWrapper interfaces

ok

nxt3d:
I rebased my changes on the latest master branch, updated isWrapped to be internal, and took out the Resolver.sol interface updates. I will push that as a separate PR. @jefflau

jefflau:
@nxt3d Can you add `upgrade()` to the interface as well? https://github.com/ensdomains/ens-contracts/blob/master/contracts/wrapper/NameWrapper.sol#L481

nxt3d:
> @nxt3d Can you add `upgrade()` to the interface as well?

When I rebased, upgrade was added to the INameWrapper, so it's there now.

jefflau:
Can we move `isWrapped()` to the internal section of the contracts and prefix it with an underscore. I don't think we need the comment since it's no longer public

jefflau:
I've merged and I'll make the updates in a separate branch

PR #209, opened 2023-02-20
Fix UniversalResolver CCIP-read issues

TateB:
issues fixed:
- `_resolveCallback()` was incorrectly using the loop index as a reference when checking for an offchain lookup failure, instead of the offchain lookup index. this lead to occasionally out-of-bounds solidity errors.
- responses from resolvers that don't implement the `ExtendedResolver` interface would be incorrectly unwrapped from `bytes`, which lead to errors.


Arachnid (approved):

contracts/utils/UniversalResolver.sol
```diff
@@ -426,32 +416,31 @@ contract UniversalResolver is ERC165, Ownable {
         returns (
```
jefflau, line 416: Add result to natspec

jefflau (changes requested):

jefflau (approved):

PR #210, opened 2023-02-20
updated Resolver.sol to include approvals

nxt3d:


nxt3d:
Maybe we should add tests for Resolver.sol? I didn't see any tests to update when I added the functions. @jefflau

jefflau:
Happy for some tests to be written. I'll merge this in first so it can be used.

PR #212, opened 2023-02-21
Fix namewrapper deploy

TateB:
`NameWrapper` deploy script was using `INameWrapper` as the interface but there are multiple instances of `INameWrapper`, switched to fully qualified name so it knows which one to use


jefflau:
Closing as was fixed by Premm's other PR

PR #217, opened 2023-02-23
Add ERC721 approve

jefflau:
**Code changes**
- Adds ERC721 Approval
- `getApproved` instead of reverting when ownerOf()` is `address(0)` returns `address(0)`.
- Approved owner cannot transfer, but has all other powers to change a name.
- Clears approved when transfer, the same as in ERC721, but does not clear when `CANNOT_APPROVE` has been burnt
- If `CANNOT_APPROVE` has been burnt, approve reverts

**Possible features**
- Allows unruggable renewals for rentable subdomains as you can approve a contract and then revoke the ability to change approvals using fuses
- Allows guaranteed registrations for a period of time
- Allows setting something similar to the .eth registrar with controllers without having to transfer the name to a contract.


Arachnid (approved):

nxt3d:
I suggest that we change the behavior of CAN_EXTEND_EXPIRY to instead of allowing the owner to renew the name, allow the approved address of the subname to renew the name. This will allow for both unruggable rentals and sales, because it is also possible to have the owner as the approved address as well as a "renewal" controller.

nxt3d:
> 
> then either the owner of the name _or_ any approved address can extend it.
> 
The idea is to make it so either the owner can extend or the approved address can extend but not both.

If we just use the approved address for CAN_EXTEND_EXPIRY then it can either be set to the owner or an approved "renewal contract". Because we can lock the approved address this works perfect for both rentals or sales.

jefflau:
Why would we want to restrict the owner from extending?

nxt3d:
> Why would we want to restrict the owner from extending?

For rentals we need to add conditions for renewing the name, namely payment with ETH or a token.

jefflau:
> > Why would we want to restrict the owner from extending?
> 
> For rentals we need to add conditions for renewing the name, namely payment with ETH or a token.

That's done by the approved contract. I don't see the reason why you need to restrict the owner from renewing.

contracts/wrapper/NameWrapper.sol
```diff
@@ -203,8 +265,27 @@ contract NameWrapper is
...
+
+    function canModifySubname(
+        bytes32 node,
+        address addr
+    ) public view returns (bool) {
+        (address owner, uint32 fuses, uint64 expiry) = getData(uint256(node));
+        return
+            (owner == addr ||
+                isApprovedForAll(owner, addr) ||
+                (owner != address(0) &&
```
Arachnid, line 286: Can't this just call `getApproved`, since the same check happens there?

contracts/wrapper/NameWrapper.sol
```diff
@@ -1096,4 +1182,12 @@ contract NameWrapper is
             ownerOf(uint256(node)) != address(0) &&
             ens.owner(node) == address(this);
     }
+
+    function _isDotEthAndIsNotExpired(
```
Arachnid, line 1186: Shouldn't this be something like `isNotDotEthOrIsNotExpired`? Can you elaborate on why we don't need to check expirations for non-.eth names?

Arachnid (changes requested):

contracts/wrapper/NameWrapper.sol
```diff
@@ -442,7 +523,7 @@ contract NameWrapper is
         bytes32 node = _makeNode(parentNode, labelhash);
 
         // this flag is used later, when checking fuses
-        bool canModifyParentName = canModifyName(parentNode, msg.sender);
+        bool canModifyParentName = canModifySubname(parentNode, msg.sender);
```
Arachnid, line 526: This variable probably needs renaming.

contracts/wrapper/NameWrapper.sol
```diff
@@ -856,6 +937,11 @@ contract NameWrapper is
                 revert OperationProhibited(bytes32(id));
             }
         }
+
+        // delete token approval if CANNOT_APPROVE has not been burnt
+        if (fuses & CANNOT_APPROVE == 0) {
```
Arachnid, line 942: Can we do this somewhere other than in a check?

contracts/wrapper/NameWrapper.sol
jefflau: This is probably a bad name. The reason this check exists is because we wanted to set the expiry to .eth expiry + grace period. However we do not want to allow them to keep operating on the name, so for the canModifyName checks we subtracted GRACE_PERIOD from the expiry.

contracts/wrapper/NameWrapper.sol
jefflau: I'm not sure we can put it anywhere else, since safeBatchTransferFrom doesnt use `_transfer` (which is the only other place I thought we could put it, but safeBatchTransferFrom would also need to change this fuse. We could rename so it doesn't use the word check?

contracts/wrapper/NameWrapper.sol
Arachnid: The name definitely needs changing; it doesn't do what it says right now.

Doesn't this check still allow operating on expired names as long as they're not .eth 2LDs?

contracts/wrapper/NameWrapper.sol
Arachnid: That would at least be an improvement.

contracts/wrapper/NameWrapper.sol
```diff
@@ -203,8 +265,26 @@ contract NameWrapper is
...
+    }
+
+    /**
+     * @notice Checks if owner/operator or approved by owner
+     * @param node namehash of the name to check
+     * @param addr which address to check permissions for
+     * @return whether or not is owner/operator or approved
+     */
+
+    function canModifySubname(
```
Arachnid, line 278:
```suggestion
    function canModifySubnames(
```

contracts/wrapper/NameWrapper.sol
```diff
@@ -102,6 +103,54 @@ contract NameWrapper is
         return super.ownerOf(id);
     }
 
+    function safeTransferFrom(
```
Arachnid, line 106: What made overriding this necessary?

contracts/wrapper/NameWrapper.sol
```diff
@@ -1096,4 +1181,12 @@ contract NameWrapper is
             ownerOf(uint256(node)) != address(0) &&
             ens.owner(node) == address(this);
     }
+
+    function _isNotDotEthOrIsNotExpired(
```
Arachnid, line 1185: I don't follow the logic here. Why do we restrict modifying the subname to non .eth 2LDs that aren't expired? And why do we apply the grace period to non .eth 2LDs? Don't we only have the grace period on .eth 2LDs?

contracts/wrapper/NameWrapper.sol
jefflau: No it doesn't. Because when a name is truly expired, the owner is zero'd out by getData(). However .eth are special now because their expiry now includes the grace period

contracts/wrapper/NameWrapper.sol
jefflau: This is something you yourself put in when we added the grace period to expiry. If it's not a .eth, we just ignore and return true. We don't apply grace period to non .eth 2LDs, if it's not a .eth the statement just returns true.

contracts/wrapper/NameWrapper.sol
jefflau: Changed in a separate comment

contracts/wrapper/NameWrapper.sol
jefflau: This was for a bug that allowed .eth names to be transferred during grace period

Arachnid (approved):

PR #232, opened 2023-03-14
Allow the "approved address" to call extendExpiry instead of the owner when CAN_EXTEND_EXPIRY is burned. 

nxt3d:
**Why is it necessary to have a “Renewal Controller” for every subname?**

With a single parent level renewal controller (approved contract), it is necessary to predict exactly how all names in the namespace will be used, which can be difficult or even impossible. However, with approved contracts "renewal controllers" attached to each individual subname, it is possible to start using the namespace right away. For instance, one could create a contract for a numbers collection, e.g. 247.netdao.eth,  and start issuing names, then move on to a membership contract and begin using member names, e.g. tom.netdao.eth. An event can be created and tickets sold with an event renewal controller, and later on, a renewal controller can be made for a metaverse platform, and so on.

**Is it still possible for the owner to extend their own subname?**
Yes, the owner can be set as the approved address for the subname, or a contract can be approved that allows the owner to renew their own subname.

Note: We are getting very close to the max contract size. In order to make the tests pass I deleted ERC20Recoverable. It is possible with some work to get it back by doing some refactoring, in the case where this PR is adopted.

### DIFF
--- a/contracts/wrapper/ERC1155Fuse.sol
+++ b/contracts/wrapper/ERC1155Fuse.sol
@@ -40,7 +40,6 @@ abstract contract ERC1155Fuse is ERC165, IERC1155, IERC1155MetadataURI {
      */
     function approve(address to, uint256 tokenId) public virtual {
         address owner = ownerOf(tokenId);
-        require(to != owner, "ERC721: approval to current owner");
 
         require(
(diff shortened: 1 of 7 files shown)


jefflau:
NameWrapper has been deployed to mainnet so closing this PR

PR #237, opened 2023-03-29
fix: add wildcard support check in universal resolver

TateB:
according to ENSIP-10, if using a resolver as a wildcard resolver it needs to support `resolve()`, otherwise it should revert.

### DIFF
--- a/contracts/utils/UniversalResolver.sol
+++ b/contracts/utils/UniversalResolver.sol
@@ -524,7 +524,7 @@ contract UniversalResolver is ERC165, Ownable {
 
         require(
             !multicallData.isWildcard || hasExtendedResolver,
-            "UniversalResolver: Wildcard on legacy resolvers is not supported"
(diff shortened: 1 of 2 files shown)


contracts/utils/UniversalResolver.sol
```diff
@@ -510,6 +522,11 @@ contract UniversalResolver is ERC165, Ownable {
         bool isCallback = multicallData.name.length == 0;
         bool hasExtendedResolver = _hasExtendedResolver(multicallData.resolver);
 
+        require(
+            !multicallData.isWildcard || hasExtendedResolver,
+            "UniversalResolver: Wildcard on legacy resolvers is not supported"
```
jefflau, line 527: Maybe something like "UniversalResolver: Wildcard on non-extended resolvers is not supported"

jefflau (changes requested):

jefflau (approved):

PR #247, opened 2023-04-28
Feature/Allow signed messages to claim reverse nodes

jefflau:
* Allows an address to assign a relayer to claim on its behalf
* Would allow contract wallets to claim their reverse node after deploy as it uses ERC-1271 method of verification
* Uses expiry instead of a nonce. However possible attack is to trick a user into signing a message far into the future which would allow them to change the reverse node possibly indefinitely. Have added upper limit to expiry. May consider moving back to a nonce for replay attack protection.

### DIFF
--- a/contracts/reverseRegistrar/ReverseRegistrar.sol
+++ b/contracts/reverseRegistrar/ReverseRegistrar.sol
@@ -132,6 +132,7 @@ contract ReverseRegistrar is Ownable, Controllable, IReverseRegistrar {
                 signatureExpiry
             )
         );
+
         bytes32 message = hash.toEthSignedMessageHash();
 
         if (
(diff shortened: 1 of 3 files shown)


contracts/reverseRegistrar/ReverseRegistrar.sol
```diff
@@ -13,11 +15,20 @@ bytes32 constant lookup = 0x3031323334353637383961626364656600000000000000000000
 
 bytes32 constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
 
+bytes4 constant claimForAddrWithSignatureFuncId = bytes4(
```
Arachnid, line 18: You should be able to do this by declaring an interface, then using `interface.function.selector`.

Arachnid (approved):

PR #251, opened 2023-05-05
Fix controller event

jefflau:
Fixes the renew event to emit the correct price


Arachnid (approved):
Please merge to a new 'dev' branch.

PR #253, opened 2023-05-16
fix: set_tlds deploy script

TateB:
`set_tlds` deploy script was broken. it wasn't picked up by the github action because it was being skipped.
changes:
- added tlds array for hardhat network (so that it runs on github actions)
- added dns encoded tld arg for the `enableNode()` function
- removed `Root` dep


jefflau (approved):

makoto:
I am getting `Error: Cannot find module 'idna-uts46-hx/uts46bundle.js'`. Can you add the package?

PR #254, opened 2023-05-23
Update deploy scripts in preparation for testnet deploy

jefflau:
* Added OwnedResolver script
* Added sepolia config
* Added setting eth resolver to owned resolver in controller

### DIFF
--- a/deploy/resolvers/00_deploy_eth_owned_resolver.ts
+++ b/deploy/resolvers/00_deploy_eth_owned_resolver.ts
@@ -15,8 +15,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   if (!ethOwnedResolver.newlyDeployed) return
 }
 
-func.id = 'resolver'
-func.tags = ['resolvers', 'EthOwnedResolver']
+func.id = 'owned-resolver'
+func.tags = ['resolvers', 'OwnedResolver']
 func.dependencies = []
 
 export default func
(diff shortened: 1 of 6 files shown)


deploy/resolvers/00_deploy_eth_owned_resolver.ts
```diff
@@ -0,0 +1,22 @@
...
+  const deployArgs = {
+    from: deployer,
+    args: [],
+    log: true,
+  }
+  const ethOwnedResolver = await deploy('OwnedResolver', deployArgs)
+  if (!ethOwnedResolver.newlyDeployed) return
+}
+
+func.id = 'resolver'
```
Arachnid: This needs a more specific name, like 'OwnedResolver'.

deploy/resolvers/00_deploy_eth_owned_resolver.ts
```diff
@@ -0,0 +1,22 @@
...
+func.tags = ['resolvers', 'EthOwnedResolver']
```
Arachnid: I've been using the convention of making one of the tags the contract name - so just `OwnedResolver`.

Arachnid (approved):

deploy/resolvers/00_deploy_eth_owned_resolver.ts
jefflau: So we should have 3 tags? `['resolvers', 'EthOwnedResolver, 'OwnedResolver']`? or just switch out `EthOwnedResolver` with `OwnedResolver`?

Arachnid: Unless there's a compelling need for `EthOwnedResolver`, I would replace it.

jefflau: Originally it was because I actually setup the eth resolver in this file, but moved the setup to the controller because I had issues with stackoverflow setting it up here.

PR #259, opened 2023-06-12
fix set error reverse record owner bug

ququzone:


Arachnid:
What error do you think this is fixing? The fix would allow anyone to set other peoples' reverse records.

ququzone:
> What error do you think this is fixing? The fix would allow anyone to set other peoples' reverse records.

When using `register` to register a name and the `owner` parameter value is different from `msg.sender`, the reverse record to this name will be set to `msg.sender`. I think should be set to `owner`.

Arachnid:
Closing this as it would allow people to set others' reverse records without permission.

PR #262, opened 2023-06-22
Lockable resolver

serenae-fansubs:
## Lock specific record types
* `lockABI(node)`
* `lockAddr(node)`
* `lockContenthash(node)`
* `lockDNS(node)`
* `lockInterface(node)`
* `lockName(node)`
* `lockPubkey(node)`
* `lockText(node)`

## Lock specific text record keys but not others
* `lockText(node, 'avatar')`
* `lockText(node, 'url')`

## Lock everything
* `lockAll(node)`

Once a specific record type (or everything) is locked, cannot be unlocked. Once at least one lock is performed, that node will no longer be clearable (cannot increment version).

This can be used in conjunction with the `CANNOT_SET_RESOLVER` fuse on wrapped names to create an immutable name that is guaranteed to resolve to records forever (or at least until the name expires).


Arachnid:
Is this ready for review?

jefflau:
From a high level, should we consider adding a function or functions that will first check if is a) wrapped, and b) has burnt `CANNOT_SET_RESOLVER` and then c) the record type is locked. This could then be used as by clients to know if a record is truly immutable, or it's just the resolver that has the record locked

serenae-fansubs:
> From a high level, should we consider adding a function or functions that will first check if is a) wrapped, and b) has burnt `CANNOT_SET_RESOLVER` and then c) the record type is locked. This could then be used as by clients to know if a record is truly immutable, or it's just the resolver that has the record locked

That makes sense, I can add something like that.

Also, another consideration is the expiry of the name, and whether that should clear any locks. That would increase the complexity of this contract though.

As it stands now, if this was added into the Public Resolver, then any locks for a particular node would be permanent for that particular resolver deployment. So if Owner A locks the ETH address, but then the name expires and someone re-registers the name, Owner B would not be able to change the ETH address (that is of course, the whole point). So Owner B would be forced to update to a different resolver first.

adraffy:
[PermissionedResolver](https://github.com/ensdomains/contracts-v2/blob/main/contracts/src/resolver/PermissionedResolver.sol) in ENSv2 can lock any value by revoking all associated roles.

Stale locks are avoided by the change from `PublicResolver` &rarr; `OwnedResolver`.

PR #265, opened 2023-07-24
Add L2 reverse registrar contract

jefflau:
* Resolver profiles for setName and setText
* Signature record setting for setName
* Signature record setting for setName for an Ownable contract
* Signature record setting for setText
* Signature record setting for setText for an Ownable contract
* Signature record setting for clearRecords


Zawana969 (approved):

.vscode/settings.json
Arachnid, line 1: This should probably be added to .gitignore.

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity >=0.8.4;
+
+interface IL2ReverseRegistrar {
+    function setName(string memory name) external returns (bytes32);
+
+    function setNameForAddrWithSignature(
+        address addr,
+        string memory name,
+        address relayer,
+        uint256 signatureExpiry,
```
Arachnid: I wonder if, rather than using expiry, we should use inception? That gives us an ordering over signatures and we know that we can ignore earlier ones if we have a later one.

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity >=0.8.4;
+
+interface IL2ReverseRegistrar {
+    function setName(string memory name) external returns (bytes32);
+
+    function setNameForAddrWithSignature(
+        address addr,
+        string memory name,
+        address relayer,
```
Arachnid: Why does `relayer` need to be specified here?

contracts/reverseRegistrar/L2ReverseClaimer.sol
```diff
@@ -0,0 +1,21 @@
...
+contract L2ReverseClaimer {
+    constructor(
+        address l2ReverseRegistrarAddr,
+        ENS reverseRegistrar,
+        address claimant
+    ) {
+        IL2ReverseRegistrar reverseRegistrar = IL2ReverseRegistrar(
+            l2ReverseRegistrarAddr
+        );
+        //reverseRegistrar.setName(claimant);
```
Arachnid: Should this be uncommented, or deleted?

contracts/reverseRegistrar/profiles/L2NameResolver.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.4;
+
+import "./L2ReverseResolverBase.sol";
+import "../../resolvers/profiles/INameResolver.sol";
+
+abstract contract L2NameResolver is INameResolver, L2ReverseResolverBase {
```
Arachnid: Is this different to `NameResolver`? Can't we avoid the need for duplication here by implementing the L2-specific functionality as a mixin at the top of the inheritance hierarchy, instead of the bottom?

Arachnid (changes requested):

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
jefflau: I was thinking if someone wanted to be more restrictive on who could send the tx, but maybe it doesn't matter given that all of the arguments are set within the signature.

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
jefflau: What do you mean by inception? Do you mean a nonce-based system?

contracts/reverseRegistrar/profiles/L2NameResolver.sol
jefflau: So the problem I had was, I needed access to the address not the node in authorisation. And the original resolver uses `bytes32 node`. Whereas the reverse registrar passes through address for authorisation. I wasn't sure how to change the modifier here without duplicating the contracts.

Arachnid: Given we call `_setName` in the parent contract, is there any point in making these separate contracts at all, then?

Arachnid (changes requested):

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    function setNameForAddrWithSignature(
+        address addr,
+        string memory name,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) external returns (bytes32);
+
+    function setNameForAddrWithSignatureAndOwnable(
+        address contractAddr,
```
makoto, line 19: I think it reads better if the argument order is in line with the function name then all the rest (name, adddr, signature, owner, contractAddress, inceptionDate...). Same for other functions

contracts/reverseRegistrar/L2ReverseClaimer.sol
makoto: This contract doesn't seem to be used from any other contracts nor any tests exist. Is the contract fully built out?

contracts/reverseRegistrar/L2ReverseClaimer.sol
```diff
@@ -0,0 +1,21 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.17 <0.9.0;
+
+import {ENS} from "../registry/ENS.sol";
+import {IL2ReverseRegistrar} from "../reverseRegistrar/IL2ReverseRegistrar.sol";
+
+contract L2ReverseClaimer {
+    constructor(
+        address l2ReverseRegistrarAddr,
+        ENS reverseRegistrar,
```
makoto: Isn't this supposed to be registry? Why you pass this as constructor but immediately assigning an instance at line 13?

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
```diff
@@ -0,0 +1,55 @@
+pragma solidity >=0.8.4;
+
+interface IL2ReverseRegistrar {
+    function setName(string memory name) external returns (bytes32);
```
makoto, line 4: Why do we need to return `bytes32` for transactions? `setName` on L1 doesn't return anything https://github.com/ensdomains/ens-contracts/blob/staging/contracts/resolvers/profiles/NameResolver.sol#L15 . The same applies to all `set*` functions.

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
```diff
@@ -0,0 +1,73 @@
...
+
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "../../resolvers/profiles/IVersionableResolver.sol";
+
+abstract contract L2ReverseResolverBase is ERC165 {
+    mapping(bytes32 => uint64) internal recordVersions;
+    event VersionChanged(bytes32 indexed node, uint64 newVersion);
+    bytes32 public immutable L2_REVERSE_NODE;
+
+    bytes32 constant lookup =
```
makoto: Can you add a comment explaining what `lookup` is?

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
```diff
@@ -0,0 +1,73 @@
...
+        require(isAuthorised(addr));
+        _;
+    }
+
+    /**
+     * Increments the record version associated with an ENS node.
+     * May only be called by the owner of that node in the ENS registry.
+     * @param addr The node to update.
+     */
+    function clearRecords(address addr) public virtual authorised(addr) {
```
makoto: Can you add test for this function as the implementation defers a lot from the L1 equivalent.

makoto:
What is the rational of the reverse registrar having `setText` which wasn't the case for L1 reverse registrar? Also, if there is a valid reason for it, what's the reason of reverse registrar not having `setAddr` either?

contracts/reverseRegistrar/README.md
```diff
@@ -0,0 +1,9 @@
+# L2 Reverse Registrar
+
+## Summary
+
+The L2 Reverse registrar is a combination of a resolver and a reverse registrar that allows the name to be set for a particular reverse node.
+
+## Resolver Profiles
+
+Profiles are separate from resolvers as the authorised modifier needs the address, rather than the node
```
makoto: Can you elaborate a bit more? I don't fully understand what it means. Also, isn't ETH and other addresses part of profile?

test/reverseRegistrar/TestL2ReverseRegistrar.js
```diff
@@ -0,0 +1,289 @@
...
+  before(async function () {
+    signers = await ethers.getSigners()
+    account = await signers[0].getAddress()
+    account2 = await signers[1].getAddress()
+
+    const L2ReverseRegistrarFactory = await ethers.getContractFactory(
+      'L2ReverseRegistrar',
+    )
+    L2ReverseRegistrar = await L2ReverseRegistrarFactory.deploy(
+      namehash('optimism.reverse'),
```
makoto, line 34: Are we using `optimism.reverse` instead of cointype?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,329 @@
...
+                name,
+                inceptionDate
+            )
+        );
+
+        bytes32 message = hash.toEthSignedMessageHash();
+
+        if (
+            !SignatureChecker.isValidSignatureNow(addr, message, signature) ||
+            inceptionDate <= lastUpdated[node] ||
```
makoto: Is this check necessary assuming `lastUpdated[node]` is always > `block.timestamp` ?

test/reverseRegistrar/TestL2ReverseRegistrar.js
```diff
@@ -0,0 +1,289 @@
...
+  })
+
+  describe('setNameForAddrWithSignature', () => {
+    it('allows an account to sign a message to allow a relayer to claim the address', async () => {
+      const funcId = ethers.utils
+        .id(setNameForAddrWithSignatureFuncSig)
+        .substring(0, 10)
+
+      const block = await ethers.provider.getBlock('latest')
+      const signatureExpiry = block.timestamp + 3600
```
makoto: Isn't this supposed to be `inceptionDate` (and you need a test case to check the failure scenario when `inceptionDate = block.timestamp - 1`)

makoto: Or rather you should have both inceptionDate and signatureExpiry support

makoto:
More of a general observation but do they have any L2 specific traits about these contracts? The majority of the work is to allow registration by signature which is useful on L1 as well. Can we make this as a default name registrar on L1 as well?

contracts/reverseRegistrar/profiles/L2TextResolver.sol
```diff
@@ -0,0 +1,39 @@
...
+abstract contract L2TextResolver is ITextResolver, L2ReverseResolverBase {
+    mapping(uint64 => mapping(bytes32 => mapping(string => string))) versionable_texts;
+
+    function _setText(
+        bytes32 node,
+        string calldata key,
+        string calldata value
+    ) internal {
+        versionable_texts[recordVersions[node]][node][key] = value;
+        emit TextChanged(node, key, key, value);
```
makoto: Because L2 doesn't have a registry, registrar nor namewrapper, we don't have a way to decode the node when indexing. dm3's resolver changed [the behavior to pass DNS encoded name] (https://github.com/corpus-io/ENS-Bedrock-Resolver/blob/main/contracts/l2/profiles/text/TextResolver.sol#L19). Is it worth changing the interfaces for all `set*` functions to take `bytes calldata name`, instead of `bytes32 node` ? cc @Arachnid

makoto:
You probably need to add `function supportsInterface`

contracts/reverseRegistrar/L2ReverseClaimer.sol
jefflau: Removed for now, we can possibly add it again later

contracts/reverseRegistrar/L2ReverseRegistrar.sol
jefflau: Wouldn't `lastUpdated[node]` always be < `block.timestamp`

test/reverseRegistrar/TestL2ReverseRegistrar.js
jefflau: No, but it was just an example since we it's a test.

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
jefflau: I'm actually not 100% sure, this is lifted from the original reverse registar

contracts/reverseRegistrar/L2ReverseRegistrar.sol
jefflau: I think this has changed since and I've added comments, so maybe worth looking at again

jefflau:
> You probably need to add `function supportsInterface`

I think there is already a function there?

test/reverseRegistrar/TestL2ReverseRegistrar.js
```diff
@@ -241,6 +287,57 @@ describe('L2ReverseRegistrar', function () {
         ),
       ).to.be.revertedWith(`InvalidSignature()`)
     })
+
+    it('reverts if inceptionDate is too low', async () => {
```
makoto: a bit pedantic but this test says `inceptionDate` while the other test for `setNameForAddrWithSignature ` says `inception date`. Probably better to standardise to either

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
makoto: The implementation code of `clearRecords` and `clearRecordsWithSignature` are identical (only difference in modifier). Can you refactor to avoid code duplication?

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
makoto: @Arachnid do you know what this is?

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
makoto: This point still stands

jefflau: The original reverseRegistrar does this: https://github.com/ensdomains/ens-contracts/blob/staging/contracts/reverseRegistrar/IReverseRegistrar.sol

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
jefflau: Normally we pass what the identifier is first, e.g. for ENS names it's their namehash, similarly the identifier for the what records is being edited is the address, not the name. So IMO this should be the address

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+    ITextResolver,
+    INameResolver,
+    IL2ReverseRegistrar
+{
+    using ECDSA for bytes32;
+    mapping(bytes32 => uint256) public lastUpdated;
+    mapping(uint64 => mapping(bytes32 => mapping(string => string))) versionable_texts;
+    mapping(uint64 => mapping(bytes32 => string)) versionable_names;
+    mapping(bytes32 => uint64) internal recordVersions;
+    event VersionChanged(bytes32 indexed node, uint64 newVersion);
```
Arachnid, line 35: Can you put all the events together?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+    bytes32 public immutable L2_REVERSE_NODE;
+
+    bytes32 constant lookup =
```
Arachnid, line 43: A comment, or a more descriptive variable name, would be good.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+        0x3031323334353637383961626364656600000000000000000000000000000000;
+
+    event ReverseClaimed(address indexed addr, bytes32 indexed node);
+
+    /**
+     * @dev Constructor
+     */
+    constructor(bytes32 L2ReverseNode) {
+        L2_REVERSE_NODE = L2ReverseNode;
```
Arachnid: We should reserve all-caps for constants.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) {
+        isAuthorisedWithSignature(hash, addr, inceptionDate, signature);
+        _;
+    }
+
+    function isAuthorised(address addr) internal view returns (bool) {
+        require(
```
Arachnid: We should use custom errors here.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+    function isAuthorisedWithSignature(
+        bytes32 hash,
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) internal view returns (bool) {
+        bytes32 message = hash.toEthSignedMessageHash();
+        bytes32 node = _getNamehash(addr);
+
+        if (
```
Arachnid, line 107: Can we give separate errors for an invalid signature and an out-of-date one?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    function isAuthorised(address addr) internal view returns (bool) {
+        require(
+            addr == msg.sender || ownsContract(addr, msg.sender),
+            "ReverseRegistrar: Caller is not a controller or authorised by address or the address itself"
+        );
+    }
+
+    function isAuthorisedWithSignature(
```
Arachnid, line 92: We need to include the chain ID in the signed message - otherwise, the same message could be replayed on multiple chains.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+    )
+        public
+        override
+        authorisedSignature(
+            keccak256(
+                abi.encodePacked(
+                    IL2ReverseRegistrar.setNameForAddrWithSignature.selector,
+                    addr,
+                    name,
+                    inceptionDate
```
Arachnid: Why not make this part of an outer hashing operation, so it doesn't have to be encoded in the inner signature *and* passed as a separate parameter?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+     * @return The ENS node hash of the reverse record.
+     */
+    function setNameForAddrWithSignatureAndOwnable(
+        address contractAddr,
+        address owner,
+        string memory name,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) public returns (bytes32) {
+        bytes32 node = _getNamehash(contractAddr);
```
Arachnid, line 217: Can we either change this to use a modifier, or change the other function to not?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+                contractAddr,
+                owner,
+                name,
+                inceptionDate
+            )
+        );
+
+        bytes32 message = hash.toEthSignedMessageHash();
+
+        if (
```
Arachnid: Why not use `isAuthorisedWithSignature` here?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+        address contractAddr,
+        address owner,
+        string calldata key,
+        string calldata value,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) public returns (bytes32) {
+        bytes32 node = _getNamehash(contractAddr);
+
+        bytes32 hash = keccak256(
```
Arachnid: As above RE modifiers etc.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) internal view returns (bool) {
+        bytes32 message = hash.toEthSignedMessageHash();
+        bytes32 node = _getNamehash(addr);
+
+        if (
+            !SignatureChecker.isValidSignatureNow(addr, message, signature) ||
+            inceptionDate < lastUpdated[node] || // must be newer than current record
```
Arachnid: Thinking about this, it probably needs to be <=. The presence of `clearRecordsWithSignature` means signed messages aren't idempotent, and if there were multiple records with the same inception, an attacker could choose which (if any) of them to set. Let's make the timestamp in milliseconds, though, so it's possible to sign multiple records at once.

contracts/reverseRegistrar/README.md
```diff
@@ -0,0 +1,5 @@
+# L2 Reverse Registrar
+
+## Summary
+
+The L2 Reverse registrar is a combination of a resolver and a reverse registrar that allows the name to be set for a particular reverse node.
```
Arachnid, line 5: Some more docs would be good!

Arachnid (changes requested):

contracts/reverseRegistrar/profiles/L2ReverseResolverBase.sol
Arachnid: It's just '0123456789abcdef' in hex form; it allows quickly mapping from a number from 0-15 to a hexadecimal character.

contracts/reverseRegistrar/IL2ReverseRegistrar.sol
makoto: Isn't it the contract you added several months ago? What was the intention at that time? If it's wrong then it should be fixed now unless it requires some sort of complex migration work as the change itself is not so critical

contracts/reverseRegistrar/L2ReverseRegistrar.sol
jefflau: How could we do this? I thought about just passing through parameters and hashing inside the modifier, but it means I'd need a modifier per signature, because the signatures differ slightly per function

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+            inceptionDate >= block.timestamp // must be in the past
+        ) {
+            revert InvalidSignature();
+        }
+    }
```
Arachnid:
```suggestion
    function isAuthorisedWithSignature(
        bytes32 hash,
        address addr,
        uint256 inceptionDate,
        bytes memory signature
    ) internal view returns (bool) {
        bytes32 outerHash = keccak256(abi.encodePacked(hash, inceptionDate));
        bytes32 message = outerHash.toEthSignedMessageHash();
        bytes32 node = _getNamehash(addr);

        if (
            !SignatureChecker.isValidSignatureNow(addr, message, signature) ||
            inceptionDate < lastUpdated[node] || // must be newer than current record
            inceptionDate >= block.timestamp // must be in the past
        ) {
            revert InvalidSignature();
        }
    }
```

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,483 @@
...
+                    IL2ReverseRegistrar.setNameForAddrWithSignature.selector,
+                    addr,
+                    name,
+                    inceptionDate
+                )
+            ),
+            addr,
+            inceptionDate,
+            signature
+        )
```
Arachnid:
```suggestion
        authorisedSignature(
            keccak256(
                abi.encodePacked(
                    IL2ReverseRegistrar.setNameForAddrWithSignature.selector,
                    addr,
                    name
                )
            ),
            addr,
            inceptionDate,
            signature
        )
```

Arachnid (changes requested):

Arachnid (approved):

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -0,0 +1,519 @@
...
+    mapping(uint64 => mapping(bytes32 => string)) versionable_names;
+    mapping(bytes32 => uint64) internal recordVersions;
+    event VersionChanged(bytes32 indexed node, uint64 newVersion);
+    event ReverseClaimed(address indexed addr, bytes32 indexed node);
+
+    bytes32 public immutable L2ReverseNode;
+    uint256 public immutable coinType;
+
+    // This is the hex encoding of the string 'abcdefghijklmnopqrstuvwxyz'
+    // It is used as a constant to lookup the characters of the hex address
```
stevieraykatz, line 42: I don't think this is accurate. It's the hex encoding of the hex characters `0123456789abcdef`. 

See replication of this logic in Basenames Sha3 lib: https://github.com/base/basenames/blob/d00f71d822394cfaeab5aa7aded8225ef1292acc/src/lib/Sha3.sol#L13

PR #278, opened 2023-09-29
Deploy new DNSSEC registrar to Sepolia

jefflau:


deployments/sepolia/.pendingTransactions
Arachnid, line 1: I don't think you want to check this in.

mdtanrikulu (approved):
whoop whoop!

Screenshot 2023-10-11 at 13 34 20

PR #280, opened 2023-09-29
Add tests for OwnedResolver

### LINKED ISSUE #243: Add tests for OwnedResolver
OwnedResolver currently has no tests. We could have tests to make sure

- [ ] Test that the owner can change records
- [ ] Test that no other account can change records

subashcs:
- Adds the test cases for OwnedResolver taking reference from the PublicResolver function. 
- Updates the SPDX license for the smart contract as it was throwing a warning while compiling ( check if it is correct) 

Fixes #243

### DIFF
--- a/test/test-utils/dns.js
+++ b/test/test-utils/dns.js
@@ -7,8 +7,36 @@ function hexEncodeName(name) {
 function hexEncodeTXT(keys) {
   return '0x' + packet.answer.encode(keys).toString('hex')
 }
+// resolves to a hexadecimal address from a name
+
+function nameToHex(name) {
+  // strip leading and trailing .
+  const n = name.replace(/^\.|\.$/gm, '')
+
+  var bufLen = n === '' ? 1 : n.length + 2
+  var buf = Buffer.allocUnsafe(bufLen)
+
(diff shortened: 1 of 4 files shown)


test/test-utils/dns.js
```diff
@@ -7,8 +7,36 @@ function hexEncodeName(name) {
 function hexEncodeTXT(keys) {
   return '0x' + packet.answer.encode(keys).toString('hex')
 }
+// resolves to a hexadecimal address from a name
```
Arachnid, line 10: What's the purpose of rewriting this instead of using `dnsName`?

Arachnid (commented):
Thanks!

test/test-utils/dns.js
subashcs: Thanks for the question @Arachnid. Actually, the "dnsName" function is renamed to the "nameToHex" function.  Also, I have moved this "nameToHex" function to the utils section for sharing between the newly created TestOwnedResolver and the old TestPublicResolver.

Hari-Bombon:
@makoto @subashcs  I would like to work on this issue

PR #288, opened 2023-10-30
Delegatable resolver

makoto:
The Delegatable Reolver allows the resolver owner to add an operator to update records of a node on behalf of the owner.

It is going to be created by a factory contract every time a user set up a resolver on L2. The owner can grant operator to any names including subnames.


contracts/resolvers/DelegatableResolver.sol
```diff
@@ -0,0 +1,121 @@
...
+     * @dev Check to see if the operator has been approved by the owner for the node.
+     */
+    function getAuthorizedNode(
+        bytes memory name,
+        uint256 offset,
+        address operator
+    ) public view returns (bytes32 node, bool authorized) {
+        uint256 len = name.readUint8(offset);
+        node = bytes32(0);
+        if (len > 0) {
```
Arachnid, line 62: Doing the check for the current node before recursing would save gas.

contracts/resolvers/DelegatableResolver.sol
```diff
@@ -0,0 +1,121 @@
...
+    constructor(address owner) {
+        operators[bytes32(0)][owner] = true;
+        transferOwnership(owner);
+    }
+
+    //node => (delegate => isAuthorised)
+    mapping(bytes32 => mapping(address => bool)) operators;
+
+    /**
+     * @dev Check to see if the operator has been approved by the owner for the node.
```
Arachnid: The arguments and return types need documenting here.

Specifically, it appears that the returned `node` is always the namehash of `name`; a naive reading of the function would assume it's the namehash of the name the user is authorised to update.

test/resolvers/TestDelegatableResolver.js
```diff
@@ -0,0 +1,136 @@
...
+    it('forbids setting new address by non-owners', async () => {
+      await exceptions.expectFailure(
+        resolver.methods['setAddr(bytes32,address)'](node, operator, {
+          from: operator,
+        }),
+      )
+    })
+  })
+
+  describe('authorisations', async () => {
```
Arachnid, line 109: Can you add a test verifying that approval doesn't apply to names outside that scope (for example, approving a.b.c doesn't authorize d.b.c)?

Arachnid (changes requested):

contracts/resolvers/DelegatableResolver.sol
makoto: Not sure what you mean. `if (len > 0) ` at line 62 is happening before recursive call happens on line 64.

You mean flipping the if order like this?

```
        if(len <= 0){
            return (node, authorized || operators[node][operator]);
        } else {
            bytes32 label = name.keccak(offset + 1, len);
            (node, authorized) = getAuthorizedNode(
                name,
                offset + len + 1,
                operator
            );
            node = keccak256(abi.encodePacked(node, label));
        }
        
```

Arachnid: Never mind - I forgot that we're also using the recursion to calculate the namehash.

makoto:
PTAL @Arachnid

contracts/resolvers/DelegatableResolver.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    /*
+     * Check to see if the operator has been approved by the owner for the node.
+     * @param name The ENS node to query
+     * @param offset The offset of the label to query recursively. Start from the 0 position and kepp adding the length of each label as it traverse. The function exits when len is 0.
+     * @param operator The address of the operator to query
+     * @return node The node of the name passed as an argument
+     * @return authorized The boolean state of whether the operator is approved to update record of the name
+     */
+    function getAuthorizedNode(
```
jefflau: nit, `isAuthorised()` is british spelling, but `getAuthorizedNode()` is american. We should pick one for this contract, since most of ENS already has british spelling, I'd stick with that!

contracts/resolvers/DelegatableResolver.sol
```diff
@@ -0,0 +1,126 @@
...
+        bytes32 indexed node,
+        address indexed operator,
+        bytes name,
+        bool approved
+    );
+
+    error NotAuthorized(bytes32 node);
+
+    constructor(address owner) {
+        operators[bytes32(0)][owner] = true;
```
jefflau: What's the purpose of this?

Arachnid: It makes the owner the owner of the root node on this resolver, which allows them to update records anywhere in the hierarchy.

Arachnid:
Is this PR now obsolete?

makoto:
> Is this PR now obsolete?

No, it is not osbolute. https://github.com/ensdomains/ens-contracts/pull/289 is to merge the branch into this branch

contracts/resolvers/DelegatableResolver.sol
makoto: This part will be replaced by the factory PR https://github.com/ensdomains/ens-contracts/pull/289

Arachnid (approved):

Chomtana:
So, we should use this branch?

PR #289, opened 2023-10-30
Delegatable resolver with factory

makoto:


contracts/resolvers/DelegatableResolver.sol
```diff
@@ -73,6 +68,12 @@ contract DelegatableResolver is
             );
             node = keccak256(abi.encodePacked(node, label));
         }
+        if (node == bytes32(0)) {
```
Arachnid:
```suggestion
        } else {
```

contracts/resolvers/DelegatableResolver.sol
```diff
@@ -96,6 +97,15 @@ contract DelegatableResolver is
         emit Approval(node, operator, name, approved);
     }
 
+    /*
+     * Returns the owner address passed set by the Factory
+     * @return address The owner address
+     */
+    function owner() public view returns (address) {
+        require(_getArgAddress(0) != address(0), "Owner is not set");
```
Arachnid: Can we omit this? It seems a waste of gas, and returning 0 is a valid response.

contracts/resolvers/DelegatableResolverFactory.sol
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A resolver factory that creates a dedicated resolver for each user
+ */
+
+contract DelegatableResolverFactory {
+    using ClonesWithImmutableArgs for address;
+
+    DelegatableResolver public implementation;
+    event newDelegatableResolver(address indexed resolver, address owner);
```
Arachnid:
```suggestion
    event NewDelegatableResolver(address resolver, address owner);
```

contracts/resolvers/DelegatableResolverFactory.sol
```diff
@@ -0,0 +1,43 @@
...
+    constructor(DelegatableResolver implementation_) {
+        implementation = implementation_;
+    }
+
+    /*
+     * Create the unique address unique to the owner
+     * @param address The address of the resolver owner
+     * @return address The address of the newly created Resolver
+     */
+    function createClone2(
```
Arachnid: The caller doesn't need to know the internal details of how resolvers are created.

```suggestion
    function create(
```

Arachnid (changes requested):
If you can update to the latest version of my PR, there's a couple of function name changes too.

contracts/resolvers/DelegatableResolver.sol
makoto: I was thinking if there's a way to prevent people interacting with a resolver not created by the factory

Arachnid: Well, either the owner will be 0, in which case they won't have permission to, or it'll throw an error because the immutable data isn't present - which is also fine.

contracts/resolvers/DelegatableResolverFactory.sol
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+contract DelegatableResolverFactory {
+    using ClonesWithImmutableArgs for address;
+
+    DelegatableResolver public implementation;
+    event newDelegatableResolver(address indexed resolver, address owner);
+
+    constructor(DelegatableResolver implementation_) {
+        implementation = implementation_;
```
jefflau: I believe convention is for the underscore to be prefixed
```suggestion
        implementation = _implementation;
```

makoto:
@jefflau @Arachnid PTAL

Arachnid (approved):

makoto:
> If you can update to the latest version of my PR, there's a couple of function name changes too.

You mean https://github.com/wighawag/clones-with-immutable-args/pull/19/commits ? It only has one commit. Did you push your change?

PR #292, opened 2023-11-30
feat: universal resolver safe calls + success var return

TateB:
changes:
- made `UniversalResolver` significantly more stable in it's return values
- `BatchGateway`'s errors are now documented within solidity. HTTP errors now use `HttpError(tuple(uint16 status, string message)[])` to differentiate them from regular solidity errors.
- single data `resolve` (`resolve(bytes,bytes)`/etc) calls will now revert if the call was unsuccessful, rather than just returning the error data.
  - errors from the name's resolver are propagated by being wrapped with `ResolverError(bytes)`. this makes it easier to identify when an error is coming from the UniversalResolver itself, or an external resolver.
  - batchgateway HTTP errors are propagated directly via `HttpError(tuple(uint16 status, string message)[])`
- batch data results from `resolve` are now `tuple(bool success, bytes returnData)`, rather than just `bytes`
- added `ResolverNotContract` error for if a name's resolver is not a contract
- external resolver calls are now all checked for safety via ERC165's `supportsInterface`
  - this is to avoid situations where in older versions of solidity, reverts would use all the remaining gas for a call
  - this was already implemented for extended resolver capability, but now applies to all calls (excl. callbacks)
  - unsafe calls will still be attempted with a maximum of 50k gas


contracts/utils/UniversalResolver.sol
```diff
@@ -46,6 +62,11 @@ struct OffchainLookupExtraData {
     bytes data;
 }
 
+struct Result {
```
Arachnid, line 65: Any reason not to roll the resolver address into this struct - or alternately, to skip the struct and return the bool as part of the return tuple?

Arachnid (changes requested):

contracts/utils/UniversalResolver.sol
TateB: the resolver address will be the same for every result returned from a multicall, but the success can potentially be different. for single calls the `Result` is unwrapped before returning `(bytes, address)`, reverting if unsuccessful.

Arachnid (approved):

PR #294, opened 2023-12-04
Minor QA things

lcfr-eth:
Add commitments(bytes32) to interface for the Registrar as I was adding this manually to mine for some UI checks in a hook.

Also

While writing some unit tests I was reviewing the ENS contracts for all possible revert() instances and came across an instance with makeCommitment that allows the resolver to be address(0) and setting reverseResolver True will revert as the resolver is address(0).

As i saw there is a check for resolver != address(0) when data is supplied I thought it should be considered a similar/same kind of check.

A flawed frontend can submit commitments which can never register etc.


Arachnid (approved):
Thanks!

PR #295, opened 2023-12-06
Delegatable resolver registrar

makoto:
A minimum registrar that allows the contract to approve user to update subnames.

### DIFF
--- a/contracts/resolvers/DelegatableResolverRegistrar.sol
+++ b/contracts/resolvers/DelegatableResolverRegistrar.sol
@@ -21,10 +21,14 @@ contract DelegatableResolverRegistrar is IDelegatableResolverRegistrar, ERC165 {
      */
 
     function register(bytes memory name, address operator) external {
-        bytes32 node = bytes32(0);
-        bool authorized = false;
-        (node, authorized) = resolver.getAuthorisedNode(name, 0, operator);
-        resolver.approve(name, operator, true);
+        (bytes32 node, bool authorized) = resolver.getAuthorisedNode(
+            name,
+            0,
+            operator
+        );
+        if (authorized == false) {
+            resolver.approve(name, operator, true);
+        }
     }
 
     function supportsInterface(
--- a/contracts/resolvers/DelegatableResolverRegistrar.sol
+++ b/contracts/resolvers/DelegatableResolverRegistrar.sol
@@ -1,10 +1,13 @@
+// SPDX-License-Identifier: MIT
 pragma solidity >=0.8.4;
+import "./IDelegatableResolverRegistrar.sol";
 import "./DelegatableResolver.sol";
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 
 /**
  * A delegated resolver registrar that allows anyone to register subname
  */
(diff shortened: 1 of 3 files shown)


jefflau:
Just had a thought as well. If this is being deployed per user, it could also just be the same contract as the `DelegatableResolver`. Then you wouldn't have to approve the registrar on the resolver and save a transaction and UX.

contracts/resolvers/DelegatableResolverRegistrar.sol
```diff
@@ -0,0 +1,34 @@
...
+    constructor(DelegatableResolver _resolver) {
+        resolver = _resolver;
+    }
+
+    /**
+     * @dev Approve an operator to be able to updated records on a node.
+     */
+    function register(bytes memory name, address operator) external {
+        bytes32 node = bytes32(0);
+        bool authorized = false;
```
Arachnid: You don't need to explicitly initialize these variables.

contracts/resolvers/DelegatableResolverRegistrar.sol
```diff
@@ -0,0 +1,34 @@
...
+        (node, authorized) = resolver.getAuthorisedNode(name, 0, operator);
```
Arachnid: What's the purpose of this line?

Arachnid (changes requested):
Is this intended to be used in prod, or just as an example registrar?

makoto:
`Is this intended to be used in prod, or just as an example registrar?` more as an example. I initially just had it as a gist, but had problem compiling via remix, hence decided to put into our contract. I don't mind taking this out completely from `ens-contracts` repo and put that in `evmgateway` repo instead

contracts/resolvers/DelegatableResolverRegistrar.sol
makoto: I think I just forgot to put `if(authorized == false){`

contracts/resolvers/DelegatableResolverRegistrar.sol
```diff
@@ -0,0 +1,41 @@
...
+     * @param operator  The address to approve
+     */
+
+    function register(bytes memory name, address operator) external {
+        (bytes32 node, bool authorized) = resolver.getAuthorisedNode(
+            name,
+            0,
+            operator
+        );
+        if (authorized == false) {
```
Arachnid, line 29: I don't think this does what you want. It won't prevent duplicate subdomains, because it only prevents the same person being issued the same name twice. I think instead you need to keep a mapping of issued names, or modify the DelegatableResolver to keep track of issued names.

makoto: That's my intention. I only added it to avoiding existing operator to waste the gas. Not sure restricting multiple subname operators at registrar level makes sense because the owner of the resolver contract can always approve multiple operators. If we want to change the delegatable resolver to only have single owner per subname, I think I need to add registrar feature directly in the resolver as @jefflau initially suggested.

makoto:
Going to close this as I added the contract into `evmgateway` at https://github.com/ensdomains/evmgateway/pull/23/files

PR #299, opened 2023-12-18
Feature/discount price oracle

jefflau:
* Adds a price oracle that discounts the price of a name after X years, until a max of Y years.
* Currently hardcodes these values
* Could be adjusted to be a progressive discount

Discussion here: https://discuss.ens.domains/t/a-method-to-increase-the-effective-number-of-ens-users/18440


jefflau:
Closing given that discussion on this has halted. Can reopen if discussions continue

PR #303, opened 2024-01-03
fix: universal resolver http error

TateB:
previously, UR was throwing a generic revert error where `HttpError` should be thrown because the decoding method used was incorrect. this PR fixes that by just dumb reverting the entire result (which includes the `HttpError` selector)

changes:
- switched to directly propagate the entire `HttpError` without any decoding/encoding


test/utils/TestUniversalResolver.js
```diff
@@ -560,6 +560,45 @@ contract('UniversalResolver', function (accounts) {
...
+        [true],
+        [errorData],
+      ])
+
+      try {
+        await urWithHttpErrorAbi.callStatic.resolveSingleCallback(
+          responses,
+          extraData,
+        )
+        expect(false).to.be.true
```
jefflau, line 592: `expect(false).to.be.true` this seems weird - is this supposed to be like this?

TateB: yes, we're using waffle chai matchers which are pretty outdated so afaik this is the only way to assert that a call has reverted if i want to access the data of the error

TateB: there were actually a few missing so i added those as well

jefflau (approved):

Mrt3383x38:
Binance eror lütfen

PR #318, opened 2024-01-29
FuseController

nxt3d:
Starting with SimpleController, this PR takes some of the features of NameWrapper and integrates them into a new controller called FuseController.

### DIFF
--- a/contracts/l2/FuseController.sol
+++ b/contracts/l2/FuseController.sol
@@ -80,6 +80,7 @@ contract FuseController is Ownable, IFuseController {
         require(value == 1);
         require(from == td.owner);
         require(operator == td.owner || operatorApproved);
+        require(!_isExpired(tokenData));
 
         return
             _pack(
@@ -98,6 +99,9 @@ contract FuseController is Ownable, IFuseController {
         uint256 /*id*/
     ) external view returns (uint256) {
         (address owner, , , , ) = _unpack(tokenData);
+        if (_isExpired(tokenData)) {
+            return 0;
+        }
         return _owner == owner ? 1 : 0;
     }
 
--- a/contracts/l2/FuseController.sol
+++ b/contracts/l2/FuseController.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.17;
 
 import "./L2Registry.sol";
 import "./IFuseController.sol";
-import "./IControllerUpgrade.sol";
+import "./IControllerUpgradeTarget.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 error Unauthorised(bytes32 node, address addr);
@@ -24,7 +24,7 @@ error nameExpired(bytes32 node);
 contract FuseController is Ownable, IFuseController {
     L2Registry immutable registry;
 
-    IControllerUpgrade upgradeContract;
+    IControllerUpgradeTarget upgradeContract;
 
     // A struct to hold the unpacked data
     struct TokenData {
@@ -252,7 +252,7 @@ contract FuseController is Ownable, IFuseController {
 
     // A function that sets the upgrade contract.
     function setUpgradeController(
-        IControllerUpgrade _upgradeContract
+        IControllerUpgradeTarget _upgradeContract
     ) external onlyOwner {
         upgradeContract = _upgradeContract;
     }
--- a/contracts/l2/FuseController.sol
+++ b/contracts/l2/FuseController.sol
@@ -7,6 +7,8 @@ import "./IFuseController.sol";
 import "./IControllerUpgradeTarget.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
+import "hardhat/console.sol";
+
 error Unauthorised(bytes32 node, address addr);
 error CannotUpgrade();
 error nameExpired(bytes32 node);
@@ -79,6 +81,11 @@ contract FuseController is Ownable, IFuseController {
     ) external view returns (bytes memory) {
         TokenData memory td;
 
+        // Make sure the tokenData is of the correct length.
+        if (tokenData.length < 96) {
+            revert("Invalid tokenData length");
+        }
+
         (
             td.owner,
             td.resolver,
@@ -87,11 +94,27 @@ contract FuseController is Ownable, IFuseController {
             td.renewalController
         ) = _unpack(tokenData);
 
+        require(msg.sender == address(registry), "Caller is not the registry");
         require(value == 1);
-        require(from == td.owner);
-        require(operator == td.owner || operatorApproved);
+        require(from == td.owner, "From is not the owner");
+        require(
+            operator == td.owner || operatorApproved,
+            "Operator not approved"
+        );
         (bool isExpired, , , , , ) = _isExpired(tokenData);
-        require(isExpired);
+        require(!isExpired, "Token is expired");
+
+        // Make sure the CANNOT_TRANSFER fuse is not burned.
+        require((td.fuses & CANNOT_TRANSFER) == 0, "Cannot transfer");
+
+        // if the 'to' address is the zero address, then the token is being burned, and
+        // set all the values to the default values.
+        if (to == address(0)) {
+            // Make sure the CANNOT_BURN_NAME fuse is not burned.
+            require((td.fuses & CANNOT_BURN_NAME) == 0, "Cannot burn name");
+            return
+                _pack(address(this), address(0), address(0), 0, 0, address(0));
+        }
 
         return
             _pack(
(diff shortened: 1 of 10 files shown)


contracts/l2/IController.sol
```diff
@@ -3,7 +3,11 @@
 pragma solidity ^0.8.17;
 
 interface IController {
-    function ownerOf(bytes calldata tokenData) external view returns (address);
```
Arachnid, line 6: What is the purpose of this change? Requiring controllers to implement both methods complicates them, and many will have to implement `ownerOf` simply by calling the registry back to fetch their data from it anywy.

contracts/l2/IControllerUpgrade.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.17;
+
+import "./IController.sol";
+
+interface IControllerUpgrade is IController {
```
Arachnid: Shouldn't this be something like `IUpgradeTarget`?

contracts/l2/IFuseController.sol
```diff
@@ -0,0 +1,19 @@
...
+interface IFuseController is IController {
+    function expiryOf(bytes32 node) external view returns (uint64);
+
+    function fusesOf(bytes32 node) external view returns (uint96);
+
+    function renewalControllerOf(bytes32 node) external view returns (address);
+
+    function upgrade(bytes32 node, bytes calldata extraData) external;
+
+    function setUpgradeController(
```
Arachnid, line 26: Isn't this an upgrade target, rather than any kind of controller?

contracts/l2/L2Registry.sol
```diff
@@ -1,28 +1,57 @@
 // SPDX-License-Identifier: MIT
 
 pragma solidity ^0.8.17;
+
 import "@openzeppelin/contracts/interfaces/IERC1155.sol";
 import "@openzeppelin/contracts/interfaces/IERC165.sol";
 import "@openzeppelin/contracts/utils/Create2.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
+import {IMetadataService} from "../wrapper/IMetadataService.sol";
```
Arachnid, line 10: This should definitely be its own PR. However, I think it makes more sense to generate metadata onchain, to remove the centralised ownership here.

contracts/l2/L2Registry.sol
```diff
@@ -1,28 +1,57 @@
...
-contract L2Registry is IERC1155 {
-    mapping(uint256 => bytes) public tokens;
+contract L2Registry is Ownable, IERC1155, IERC1155MetadataURI {
+    struct Record {
+        string name;
+        bytes data;
+    }
+    mapping(uint256 => Record) public tokens;
     mapping(address => mapping(address => bool)) approvals;
+    mapping(address => uint256) tokenApprovalsNonce;
```
Arachnid, line 24: This should also be in a separate PR.

contracts/l2/L2Registry.sol
```diff
@@ -186,7 +260,7 @@ contract L2Registry is IERC1155 {
             return IController(address(0));
         }
         assembly {
-            addr := shr(96, mload(add(data, 32)))
+            addr := mload(add(data, 20))
```
Arachnid, line 321: The upper bits of the address need to be masked out. I think due to that, the shr approach is probably just as efficient.

contracts/l2/FuseController.sol
```diff
@@ -0,0 +1,309 @@
...
+error nameExpired(bytes32 node);
+
+/**
+ * @dev A simple ENS registry controller. Names are permanently owned by a single account.
+ *      Name data is structured as follows:
+ *       - Byte 0: controller (address)
+ *       - Byte 20: owner (address)
+ *       - Byte 40: resolver (address)
+ *       _ Byte 60: expiry (uint64)
+ *       - Byte 68: fuses (uint96)
```
Arachnid: If you reduce this to a `uint64`, the whole thing will be 96 bytes - saving a storage slot.

contracts/l2/FuseController.sol
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    function resolverFor(
+        bytes calldata tokenData
+    ) external pure returns (address) {
+        (, address resolver, , , ) = _unpack(tokenData);
+        return resolver;
+    }
+
+    function expiryOf(bytes32 node) external view returns (uint64) {
```
Arachnid, line 196: A generic `data` method may be more useful here, as well as being more gas efficient.

contracts/l2/FuseController.sol
```diff
@@ -0,0 +1,309 @@
...
+
+        if (owner != msg.sender && !isAuthorized) {
+            revert Unauthorised(node, msg.sender);
+        }
+
+        if (_isExpired(tokenData)) {
+            revert nameExpired(node);
+        }
+
+        // Change the controller to the upgrade contract.
```
Arachnid, line 275: You may want to instead allow the upgrade contract to specify the new data, so that a data format migration could take place at the same time as a contract upgrade.

contracts/l2/FuseController.sol
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        registry.setSubnode(
+            uint256(node),
+            label,
+            _pack(
+                address(this),
+                subnodeOwner,
+                subnodeResolver,
+                subnodeExpiry,
```
Arachnid, line 587: Shouldn't there be some enforcement here that a subnode can't expire after its parent?

contracts/l2/FuseController.sol
```diff
@@ -0,0 +1,309 @@
...
+    /**********************
+     * Internal functions *
+     **********************/
+
+    function _isExpired(bytes memory tokenData) internal view returns (bool) {
+        (, , uint64 expiry, , ) = _unpack(tokenData);
+        return expiry <= block.timestamp;
+    }
+
+    function _unpack(
```
Arachnid, line 631: As I mentioned before - just use `calldata` instead of `memory` and you don't need assembly.

Arachnid (changes requested):
The fuses and expiry don't do anything yet, but I assume that's intentional?

contracts/l2/FuseController.sol
nxt3d: Using `memory` allows for the function `ownerOfWithData`, which saves an external call.

Arachnid: Why? `ownerOfWithData` can specify that its argument is `calldata` too, and pass it straight to `_unpack`.

contracts/l2/IController.sol
nxt3d: It saves an external call.

contracts/l2/L2Registry.sol
nxt3d: interesting.

contracts/l2/IController.sol
Arachnid: In what circumstances?

nxt3d:
> The fuses and expiry don't do anything yet, but I assume that's intentional?

This is for sure incomplete. I wanted to get some feedback, on the direction, to make sure we are on the same page. I will make changes, and add some more commits.

contracts/l2/FuseController.sol
nxt3d: For example in this function in the controller:

```   
function ownerOf(bytes32 node) external view returns (address) {
        //get the tokenData
        bytes memory tokenData = registry.getData(uint256(node));
        (address owner, , , , ) = _unpack(tokenData);
        return owner;
    }
```
    
 I just get the data from the registry and then call _unpack, but tokenData in this case is memory and not calldata.

contracts/l2/FuseController.sol
nxt3d: The upgrade target contract won't have permissions to make any changes if we don't change the controller here? There might be a way to make this more efficient, however.

contracts/l2/IController.sol
nxt3d:
```
function ownerOf(bytes32 node) external view returns (address) {
        //get the tokenData
        bytes memory tokenData = registry.getData(uint256(node));
        (address owner, , , , ) = _unpack(tokenData);
        return owner;
    }
```

contracts/l2/IFuseController.sol
nxt3d: setUpgradeControllerTarget? It's a controller to upgrade to.

contracts/l2/FuseController.sol
Arachnid: Right, but we could have a function on the upgrade contract that we call with the old data, and it returns the new data to set on the registry.

nxt3d: I don't have an opinion on this, could be either way.

PR #324, opened 2024-02-09
fix: universalresolver hybrid onchain/offchain handling

TateB:
previously, a batch call would add non-ENSIP10 encoded calls to`extraDatas` for onchain calls. `extraDatas` is used within `OffchainLookup`, and is directly passed through from a call to a callback, allowing the UR to re-call requests that were already onchain. since the calls in `extraDatas` were not ENSIP10 encoded, and since calls are not re-encoded in callbacks, this lead to situations where for a hybrid (onchain/offchain calls together) batch call, onchain data would be fetched on callback with the non-ENSIP10 equivalent function. 

changes:
- set data for onchain calls in `extraDatas` to be the potentially ENSIP10 encoded item.
- (misc.) improved error handling for reverse resolution failure by using the internal `_checkResolveSingle()` in `reverseCallback()`


technicallykind:
👀 Tagging myself to be notified on merge

Mrt3383x38:
Evetttttt

jefflau (approved):

PR #328, opened 2024-02-15
refactor wrapped callback

mdtanrikulu:


contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -167,6 +149,25 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
         revert CouldNotResolve(name);
     }
 
+    function resolveWrappedCallback(
+        bytes calldata response,
+        bytes calldata extraData
+    ) external view returns (bytes memory) {
+        (
+            bytes memory name,
+            bytes memory query,
```
Arachnid: Shouldn't this be called `extraData`?

Arachnid (approved):

contracts/dnsregistrar/OffchainDNSResolver.sol
mdtanrikulu: there is an argument called `extraData` which belongs to `resolveWrappedCallback()` method. 
https://github.com/ensdomains/ens-contracts/blob/9e6ebdffc3ce231777016d45b512d5bc355e124d/contracts/dnsregistrar/OffchainDNSResolver.sol#L152-L161
maybe I can call this one `innerExtraData`?

PR #331, opened 2024-02-20
Rewrite ExtendedDNSResolver to support more record types

Arachnid:
This rewrite incorporates new functionality, documented in the contract docstring, supporting address records for all coin types, as well as text records.


contracts/resolvers/profiles/ExtendedDNSResolver.sol
```diff
@@ -27,21 +74,189 @@ contract ExtendedDNSResolver is IExtendedDNSResolver, IERC165 {
...
         revert NotImplemented();
     }
+
+    function _resolveAddress(
+        bytes calldata data,
+        bytes calldata context
+    ) internal pure returns (bytes memory) {
+        (, uint256 coinType) = abi.decode(data[4:], (bytes32, uint256));
+        bytes memory value;
+        if (coinType & 0x80000000 != 0) {
```
mdtanrikulu, line 106: it may be helpful to reference the spec as a comment; https://docs.ens.domains/ensip/11#specification

mdtanrikulu (approved):
just a little comment, overall lgtm

contracts/resolvers/profiles/ExtendedDNSResolver.sol
```diff
@@ -2,18 +2,65 @@
...
+ *
+ *      For example:
+ *          ENS1 2.dnsname.ens.eth a[60]=0x1234...
+ *
+ *      The record data consists of a series of key=value pairs, separated by spaces. Keys
+ *      may have an optional argument in square brackets, and values may be either unquoted
+ *       - in which case they may not contain spaces - or single-quoted. Single quotes in
+ *      a quoted value may be backslash-escaped.
+ *
+ *      Record types:
```
makoto, line 45: How about contenthash?

Arachnid: There's no browsers that would prioritize a contenthash in ENS over the A records of the domain right now, and no reasonable prospect that that'll change any time soon.

contracts/resolvers/profiles/ExtendedDNSResolver.sol
```diff
@@ -27,21 +74,190 @@ contract ExtendedDNSResolver is IExtendedDNSResolver, IERC165 {
...
+    uint256 constant STATE_IGNORED_KEY = 1;
+    uint256 constant STATE_IGNORED_KEY_ARG = 2;
+    uint256 constant STATE_VALUE = 3;
+    uint256 constant STATE_QUOTED_VALUE = 4;
+    uint256 constant STATE_UNQUOTED_VALUE = 5;
+    uint256 constant STATE_IGNORED_VALUE = 6;
+    uint256 constant STATE_IGNORED_QUOTED_VALUE = 7;
+    uint256 constant STATE_IGNORED_UNQUOTED_VALUE = 8;
+
+    function _findValue(
```
makoto, line 170: Can you add a comment briefly describing how the logic works?

test/utils/TestHexUtils.js
```diff
@@ -98,8 +138,8 @@ describe('HexUtils', () => {
         0,
         39,
       )
-      expect(address).to.equal('0x0000000000000000000000000000000000000000')
       expect(valid).to.equal(false)
+      expect(address).to.equal('0x0000000000000000000000000000000000000000')
```
makoto: ZERO_ADDRESS

test/resolvers/TestExtendedDNSResolver.js
```diff
@@ -0,0 +1,220 @@
...
+        'text',
+        ['url'],
+        "t[url]='https://ens.domains/'",
+      )
+      expect(ethers.utils.toUtf8String(ethers.utils.arrayify(result))).to.equal(
+        'https://ens.domains/',
+      )
+    })
+
+    it('handles escaped quotes', async function () {
```
makoto, line 207: Are there any invalid texts or it literally accept any text?

test/resolvers/TestExtendedDNSResolver.js
```diff
@@ -0,0 +1,220 @@
...
+      const result = await resolve(
+        name,
+        'addr(bytes32)',
+        [],
+        `foo=bar a[60]=${testAddress}`,
+      )
+      expect(result).to.equal(testAddress.toLowerCase())
+    })
+
+    it('handles multiple spaces between records', async function () {
```
makoto, line 94: What happens if someone put comma `,` ?

test/resolvers/TestExtendedDNSResolver.js
makoto: Also does it work with any length of the string as long as DNS record can store?

contracts/resolvers/profiles/ExtendedDNSResolver.sol
aoxborrow: true, browsers won't care about any of these records, but [some people still want to associate IPFS with DNS names](https://dnslink.dev/). it would be easy to implement because it doesn't require a parameter, (`c=<hash>`), and then you have complete resolver support. 

ps. I know these records need to be evm-readable, but you have 255 chars to work with and humans involved. how about `addr[60]=0x00`, `text[avatar]='http...'`, and `contenthash=<hash>`.  maybe support `addr=0x00` and `a=0x00` as well.


pps. `dnsname.ens.eth` is a mouthful, `dns.resolver` would be nice.

Arachnid: > true, browsers won't care about any of these records, but [some people still want to associate IPFS with DNS names](https://dnslink.dev/). it would be easy to implement because it doesn't require a parameter, (c=<hash>), and then you have complete resolver support.

If there are any integrations that want to read this field from ENS names, I'll happily add support here.

> ps. I know these records need to be evm-readable, but you have 255 chars to work with and humans involved. how about addr[60]=0x00, text[avatar]='http...', and contenthash=<hash>. maybe support addr=0x00 and a=0x00 as well.

256 characters is only 5 addresses. We don't have a lot of spare space to work with here.

PR #332, opened 2024-02-20
concatenate multiple text fields

mdtanrikulu:


contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -168,6 +168,19 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
             );
     }
 
+    function stripWildcard(
+        bytes memory name
+    ) public pure returns (bytes memory) {
+        if (name.length > 4 && name[0] == "*" && name[1] == ".") {
```
Arachnid: Names here are DNS encoded, so the check should instead be `uint8(name[0]) == 1 && name[1] == '*')`.

Also, since the name is only being fetched to be compared, it'd be more efficient to write a method that checks the name using the `.equals` override that takes offsets and lengths ,rather than copying it out.

contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -199,10 +212,16 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
...
-        uint256 fieldLength = data.readUint8(startIdx);
-        assert(startIdx + fieldLength < lastIdx);
-        return data.substring(startIdx + 1, fieldLength);
+        bytes memory result = new bytes(0);
+        uint256 idx = startIdx;
+        while (idx < lastIdx) {
+            uint256 fieldLength = data.readUint8(idx);
+            assert(idx + fieldLength + 1 <= lastIdx);
+            bytes memory field = data.substring(idx + 1, fieldLength);
+            result = abi.encodePacked(result, field);
```
Arachnid: Iterated concatenation like this is `O(n^2)`. Instead, do a first loop to get the total length of the string, then allocate a new string and copy them into it in a second loop. You will probably need to write a `strcpy` method in `BytesUtils` that uses the existing `memcpy` implementation.

Arachnid (changes requested):

contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -168,17 +173,10 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
...
-        if (name.length > 4 && name[0] == "*" && name[1] == ".") {
-            bytes memory strippedName = new bytes(name.length - 2);
-            for (uint i = 2; i < name.length; i++) {
-                strippedName[i - 2] = name[i];
-            }
-            return strippedName;
-        }
-        return name;
+    ) public pure returns (bool isWildcard) {
+        return name.length > 4 && uint8(name[0]) == 1 && name[1] == "*";
```
Arachnid:
```suggestion
        return name.length => 4 && uint8(name[0]) == 1 && name[1] == "*";
```

Arachnid (changes requested):

contracts/dnsregistrar/OffchainDNSResolver.sol
mdtanrikulu: isn't minimum valid dns encoded name length 5? i.e. `\*.x.`

Arachnid: You're right, I forgot about the trailing \0!

Arachnid (approved):

contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -199,10 +200,24 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
...
-        uint256 fieldLength = data.readUint8(startIdx);
-        assert(startIdx + fieldLength < lastIdx);
-        return data.substring(startIdx + 1, fieldLength);
+        uint256 totalLength = 0;
+        uint256 idx = startIdx;
+        while (idx < lastIdx) {
+            uint256 fieldLength = data.readUint8(idx);
+            totalLength += fieldLength;
+            idx += fieldLength + 1;
+        }
```
Arachnid:
```suggestion
        for (uint256 idx = startidx; idx < lastIdx; idx += fieldLength + 1) {
            uint256 fieldLength = data.readUint8(idx);
            totalLength += fieldLength;
        }
```

contracts/dnsregistrar/OffchainDNSResolver.sol
```diff
@@ -199,10 +200,24 @@ contract OffchainDNSResolver is IExtendedResolver, IERC165 {
...
+
+        bytes memory result = new bytes(totalLength);
+        idx = startIdx;
+        uint256 resultIdx = 0;
+        while (idx < lastIdx) {
+            uint256 fieldLength = data.readUint8(idx);
+            result.strcpy(resultIdx, data, idx + 1, fieldLength);
+            resultIdx += fieldLength;
+            idx += fieldLength + 1;
+        }
```
Arachnid:
```suggestion
        uint256 resultIdx = 0;
        for (uint256 idx = startIdx; idx < lastIdx; idx += fieldLength + 1) {
            uint256 fieldLength = data.readUint8(idx);
            result.strcpy(resultIdx, data, idx + 1, fieldLength);
            resultIdx += fieldLength;
        }
```
```

Arachnid (approved):
LGTM, but a couple of suggestions to tidy the code up.

contracts/dnsregistrar/OffchainDNSResolver.sol
mdtanrikulu: made a slight change in the suggested code, since `fieldLength` variable initializes inside of the loop block;
https://github.com/ensdomains/ens-contracts/pull/332/commits/24fcb62a7f332f859e98e1e80d39770747f66792

contracts/dnsregistrar/OffchainDNSResolver.sol
mdtanrikulu: made the same change here as well;
https://github.com/ensdomains/ens-contracts/pull/332#discussion_r1844368380

Arachnid (approved):

PR #333, opened 2024-02-20
Wiki address check

LeonmanRolls:


.github/workflows/main.yml
```diff
@@ -2,10 +2,25 @@ name: CI
 
 on:
   push:
-    branches: [master]
+    branches: [staging, mainnet]
   pull_request:
 
 jobs:
+  wiki_address_chek:
```
v1rtl: typo, should be `check`

PR #335, opened 2024-02-26
Default reverse resolver

makoto:
- Add DefaultResolver
- Add SignatureResolver which is the super class of DefaultResolver and L2ReverseRegistrar
- Change signature to use eip 191 version 0


contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,264 @@
...
+    bytes32 public immutable ParentNode;
+    uint256 public immutable coinType;
+
+    /*
+     * @dev Constructor
+     * @param ParentNode The namespace to set.
+     * @param _coinType The cointype converted from the chainId of the chain this contract is deployed to.
+     */
+    constructor(bytes32 _ParentNode, uint256 _coinType) {
+        ParentNode = _ParentNode;
```
jefflau: Lowercase P

jefflau (approved):

contracts/reverseRegistrar/ISignatureReverseResolver.sol
```diff
@@ -0,0 +1,36 @@
+pragma solidity >=0.8.4;
+
+interface ISignatureReverseResolver {
+    event VersionChanged(bytes32 indexed node, uint64 newVersion);
+    event ReverseClaimed(address indexed addr, bytes32 indexed node);
```
Arachnid, line 5: This event should probably be part of `IL2ReverseRegistrar`.

contracts/reverseRegistrar/ISignatureReverseResolver.sol
```diff
@@ -0,0 +1,36 @@
...
+    event NameChanged(bytes32 indexed node, string name);
```
Arachnid, line 6: These events too.

contracts/reverseRegistrar/ISignatureReverseResolver.sol
```diff
@@ -0,0 +1,36 @@
...
+        bytes memory signature
+    ) external returns (bytes32);
+
+    function clearRecordsWithSignature(
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) external;
+
+    function node(address addr) external view returns (bytes32);
```
Arachnid: And this probably doesn't need to be part of the interface, does it?

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,264 @@
...
+        bytes32 hash,
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) {
+        isAuthorisedWithSignature(hash, addr, inceptionDate, signature);
+        _;
+    }
+
+    function getLastUpdated(
```
Arachnid: We don't use `get` as a prefix elsewhere, we probably shouldn't here.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,264 @@
...
+        bytes32 node
+    ) internal view virtual returns (uint256) {
+        return lastUpdated[node];
+    }
+
+    function isAuthorised(address addr) internal view virtual returns (bool) {
+        revert("This function needs to be overriten");
```
Arachnid:
```suggestion
        revert("This function needs to be overridden");
```

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -216,6 +132,7 @@ contract L2ReverseRegistrar is
     {
         bytes32 node = _getNamehash(contractAddr);
         _setName(node, name, inceptionDate);
+        emit NameChanged(node, name);
```
Arachnid: Shouldn't this method be part of `SignatureReverseResolver`?

Arachnid (changes requested):

contracts/reverseRegistrar/ISignatureReverseResolver.sol
makoto: They are all called within `SignatureReverseResolver` so can't be moved

contracts/reverseRegistrar/ISignatureReverseResolver.sol
makoto: `ReverseClaimed` is also emitted at `SignatureReverseResolver.setNameForAddrWithSignature` so can't move to `IL1ReverseRegistrar`.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Looks not in use so removed

makoto:
To sum up about the events, 

- `ReverseClaimed` are called by both `L2ReverseRegistrar` and `SignatureReverseResolver` (because `setNameForAddrWithSignatureAndOwnable` needs to pass `contractAddress` as the argument)
- `NameChanged` is only called by `SignatureReverseResolver`.

Hence both events should be defined at `ISignatureReverseResolver`

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -405,7 +257,7 @@ contract L2ReverseRegistrar is
     function name(
         bytes32 node
     ) external view virtual override returns (string memory) {
```
Arachnid: Is it necessary to override these at all? Couldn't we eliminate them and just rename `_name` to `name`?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -414,53 +266,7 @@ contract L2ReverseRegistrar is
...
-    }
-
-    /**
-     * @dev Returns the node hash for a given account's reverse records.
-     * @param addr The address to hash
-     * @return The ENS node hash.
-     */
-    function node(address addr) public view override returns (bytes32) {
-        return keccak256(abi.encodePacked(L2ReverseNode, sha3HexAddress(addr)));
+        _clearRecords(addr);
```
Arachnid: Ditto here.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,258 @@
...
+
+contract SignatureReverseResolver is Ownable, ISignatureReverseResolver {
+    using ECDSA for bytes32;
+    mapping(bytes32 => uint256) public lastUpdated;
+    mapping(uint64 => mapping(bytes32 => mapping(string => string))) versionable_texts;
+    mapping(uint64 => mapping(bytes32 => string)) versionable_names;
+    mapping(bytes32 => uint64) internal recordVersions;
+
+    bytes32 public immutable parentNode;
+    uint256 public immutable coinType;
```
Arachnid, line 23: We should just use chain ID here.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,258 @@
...
+        _;
+    }
+
+    function isAuthorised(address addr) internal view virtual returns (bool) {
+        revert("This function needs to be overridden");
+    }
+
+    function isAuthorisedWithSignature(
+        bytes32 hash,
+        address addr,
```
Arachnid, line 72: Why not make `addr` part of the hashed data?

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,258 @@
...
+        uint256 inceptionDate,
+        bytes memory signature
+    ) internal view returns (bool) {
+        bytes32 message = keccak256(
+            abi.encodePacked(hash, addr, inceptionDate, coinType)
```
Arachnid: This should probably follow ERC191 version 0, and include the address of the verifying contract (this address).

contracts/utils/LowLevelCallUtils.sol
```diff
@@ -67,4 +71,30 @@ library LowLevelCallUtils {
...
+                mstore8(i, byte(and(addr, 0xf), lookup))
+                addr := div(addr, 0x10)
+                i := sub(i, 1)
+                mstore8(i, byte(and(addr, 0xf), lookup))
+                addr := div(addr, 0x10)
+            }
+
+            ret := keccak256(0, 40)
+        }
+    }
```
Arachnid, line 99: Don't we already have a copy of this somewhere that we could import?

Arachnid (changes requested):

contracts/reverseRegistrar/L2ReverseRegistrar.sol
makoto: `_name(bytes32)` is used by both `L2ReverseRegistrar.name(bytes32)` and `DefaultReverseRegistrar.name(address)`. I kept `_name(bytes32)` as private because I don't want to expose the function to `DefaultReverseRegistrar.name`

contracts/reverseRegistrar/L2ReverseRegistrar.sol
makoto: Same as `_name`

contracts/utils/LowLevelCallUtils.sol
makoto: It only exists on `ReverseRegistrar` https://github.com/ensdomains/ens-contracts/blob/21736916300b26cb8ea1802dbf6c9ff054adaeab/contracts/reverseRegistrar/ReverseRegistrar.sol#L164 which I cannot import

I can refactor so that ReverseRegistrar can import the function from `LowLevelCallUtils` instead

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: `coinType` was introduced by @jefflau when he first created `isAuthorisedWithSignature` so needs his input on why `coinType` was used over `chainId`.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Same here. I just ported the function @jefflau introduced, so his input is needed.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Same here. I just ported the function @jefflau introduced, so his input is needed.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
Arachnid: Where can I see this implementation of `DefaultReverseRegistrar.name(address)`? It doesn't seem to be in this PR.

Regardless, I don't see how that precludes it calling a parent class's `name` method with a different signature, if it's defined?

contracts/utils/LowLevelCallUtils.sol
Arachnid: Yes please!

Arachnid (changes requested):

contracts/reverseRegistrar/L2ReverseRegistrar.sol
makoto: `Where can I see this implementation of DefaultReverseRegistrar.name(address)? `

It's on the evmgateway PR https://github.com/ensdomains/evmgateway/pull/33/files#diff-f685401f720e672a03e9ec861162971688a4d78c3c084741d6c1edd1f2b32e53R50

I can change to let `DefaultReverseRegistrar` have `name(address)` and `name(node)` if you say it's not too confusing

contracts/reverseRegistrar/SignatureReverseResolver.sol
jefflau: This seems reasonable to add

contracts/reverseRegistrar/SignatureReverseResolver.sol
jefflau: I think changing to chainId should be safe. I had it as cointype so we could use our derived cointype ids and possibly use this contract for non-EVM L2 chains in the future.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Instead of
`abi.encodePacked(hash, addr, inceptionDate, coinType)`
this?
`abi.encodePacked(byte(0x19), byte(0), address(this),hash, addr, inceptionDate, chainId)` @Arachnid

Arachnid: I believe that's correct, yes.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -2,68 +2,42 @@ pragma solidity >=0.8.4;
...
-    /**
+    /*
      * @dev Constructor
+     * @param _L2ReverseNode The namespace to set. The converntion is '${coinType}.reverse'
+     * @param _chainId The chainId converted from the chainId of the chain this contract is deployed to.
      */
-    constructor(bytes32 _L2ReverseNode, uint256 _coinType) {
+    constructor(
+        bytes32 _L2ReverseNode,
+        uint256 _chainId
```
makoto: I changed the constructor to pass chainId instead of cointype reflecting [your comment of another PR](https://github.com/ensdomains/evmgateway/pull/34/files#r1509124561). However, it seems a bit odd to pass chainId while reversenode still uses the coinType instead of chain Id [according to the ENSIP ](https://github.com/ensdomains/docs/pull/157/files#diff-25328e8b031181b77dc4241237ec9b26de59260f7c643435c1b632f0710d50b4R30). Should I reverse the change to use coinType instead?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -2,68 +2,42 @@ pragma solidity >=0.8.4;
...
 error NotOwnerOfContract();
 
-// @note Inception date
-// The inception date is in milliseconds, and so will be divided by 1000
-// when comparing to block.timestamp. This means that the date will be
-// rounded down to the nearest second.
-
+/**
+ * A L2 reverser registrar. Deployed to each L2 chain.
+ * The contract will be verified on L1 Reverse Resolver under the namespace specified at constructor
```
Arachnid: What does this mean?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
Arachnid: Yes, we may as well go back to using cointype.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -2,68 +2,42 @@ pragma solidity >=0.8.4;
...
+    /*
      * @dev Constructor
+     * @param _L2ReverseNode The namespace to set. The converntion is '${coinType}.reverse'
+     * @param _chainId The chainId converted from the chainId of the chain this contract is deployed to.
      */
-    constructor(bytes32 _L2ReverseNode, uint256 _coinType) {
+    constructor(
+        bytes32 _L2ReverseNode,
+        uint256 _chainId
+    ) SignatureReverseResolver(_L2ReverseNode, _chainId) {
```
Arachnid: It doesn't seem right that `L2ReverseRegistrar` should be a subclass of `SignatureReverseResolver.`. Isn't this also a resolver?

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,273 @@
...
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "../root/Controllable.sol";
+import "../utils/LowLevelCallUtils.sol";
+
+error InvalidSignature();
+error SignatureOutOfDate();
+error Unauthorised();
+
+contract SignatureReverseResolver is Ownable, ISignatureReverseResolver {
```
Arachnid: Why is this `Ownable`?

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,273 @@
...
+import "../registry/ENS.sol";
+import "./ISignatureReverseResolver.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "../root/Controllable.sol";
+import "../utils/LowLevelCallUtils.sol";
+
+error InvalidSignature();
+error SignatureOutOfDate();
```
Arachnid:
```suggestion
error InvalidSignatureDate();
```

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    function isAuthorisedWithSignature(
+        bytes32 hash,
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) internal view returns (bool) {
+        // Follow ERC191 version 0 https://eips.ethereum.org/EIPS/eip-191
+        bytes32 message = keccak256(
```
Arachnid: For convenience, we should expose functions to compute this hash, which we can then also use internally here.

contracts/reverseRegistrar/SignatureReverseResolver.sol
Arachnid: I think it should be incorporated, there's no reason for it to be a separate parameter here.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,273 @@
...
+            abi.encodePacked(
+                bytes1(0x19),
+                bytes1(0),
+                address(this),
+                hash,
+                addr,
+                inceptionDate,
+                chainId
+            )
+        ).toEthSignedMessageHash();
```
Arachnid: It didn't occur to me that we are of course using `eth_sign` here so it will work in browsers. In that case there's no point using the `bytes1(0x19), bytes1(0)` prefix, since this won't be a valid EIP191 version 0 message anyway.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,273 @@
...
+        emit VersionChanged(reverseNode, recordVersions[reverseNode]);
+    }
+
+    /**
+     * Increments the record version associated with an ENS node.
+     * May only be called by the owner of that node in the ENS registry.
+     * @param addr The node to update.
+     * @param signature A signature proving ownership of the node.
+     */
+    function clearRecordsWithSignature(
```
Arachnid, line 229: I don't think this works with signed messages; someone replaying a series of messages could choose to include or omit the `clear` signed message, producing different results either way.

Perhaps instead we can include a version in each signature message, and if the version is greater than the one stored, increment it, effectively erasing all past records?

contracts/reverseRegistrar/ReverseRegistrar.sol
```diff
@@ -86,7 +87,7 @@ contract ReverseRegistrar is Ownable, Controllable, IReverseRegistrar {
         address owner,
         address resolver
     ) public override authorised(addr) returns (bytes32) {
-        bytes32 labelHash = sha3HexAddress(addr);
+        bytes32 labelHash = LowLevelCallUtils.sha3HexAddress(addr);
```
Arachnid: If you do `using LowLevelCallUtils for address;` you can then do `addr.sha3HexAddress()`, which is much cleaner.

Arachnid (changes requested):

contracts/reverseRegistrar/L2ReverseRegistrar.sol
makoto: It just means that the l1 `address.coinType.reverse` resolver gateway fetch the data and proof from this contract and gets verified via CCIP-read. I can get rid of the sentence if it's rather confusing.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
makoto: I made as a subclass because both `L2ReverseRegistrar` and [`DefaultReverseResolver`](https://github.com/ensdomains/evmgateway/pull/33/files#diff-f685401f720e672a03e9ec861162971688a4d78c3c084741d6c1edd1f2b32e53) share the same logics via `SignatureReverseResolver`. I am happy to rename `L2ReverseRegistrar` to `L2ReverseResolver` if it makes the semantic cleaner.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Is this just a change of the order?

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Are you suggesting to just revert?

makoto:
`I think it should be incorporated, there's no reason for it to be a separate parameter here.` at https://github.com/ensdomains/ens-contracts/pull/335#discussion_r1543548879 , `addr` is used to derive `node` at `bytes32 node = _getNamehash(addr);` line 73 @Arachnid

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: I think it was simply copied from `L2ReverseRegistrar`. @jefflau does the contract have to have Ownable? If not, I can get rid of `Ownable` from both `SignatureReverseResolver` and `L2ReverseRegistrar` (All tests pass without `Ownable`)

jefflau: Fine to remove. It was an artifact from the original ReverseRegistrar

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: I am failing to understand why this is only issue for `clearRecords` but not for other `set*` functions.  Also the inclusion of the version doesn't help when the someone keeps omitting to include the `clear` signed message.
Isn't `inceptionDate` enough against most abuses?

contracts/reverseRegistrar/L2ReverseRegistrar.sol
Arachnid: It definitely needs changing or deleting.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
```diff
@@ -83,33 +55,32 @@ contract L2ReverseRegistrar is
...
-            revert SignatureOutOfDate();
-        }
+        address owner,
+        uint256 inceptionDate
+    ) public view returns (bytes32) {
+        // Follow ERC191 version 0 https://eips.ethereum.org/EIPS/eip-191
+        return
+            keccak256(
+                abi.encodePacked(
+                    bytes1(0x19),
```
Arachnid: As mentioned previously, this can be removed since we're not signing a raw ERC-191 type 0 message.

contracts/reverseRegistrar/SignatureReverseResolver.sol
Arachnid: I'm suggesting a better name for the error.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,276 @@
...
+    ) {
+        isAuthorisedWithSignature(hash, addr, inceptionDate, signature);
+        _;
+    }
+
+    function isAuthorised(address addr) internal view virtual returns (bool) {
+        revert("This function needs to be overridden");
+    }
+
+    function computeMessage(
```
Arachnid: This should be called something like `getSignedMessageHash`.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,276 @@
...
+        bytes32 hash,
+        address addr,
+        uint256 inceptionDate
+    ) public view returns (bytes32) {
+        // Follow ERC191 version 0 https://eips.ethereum.org/EIPS/eip-191
+        return
+            keccak256(
+                abi.encodePacked(
+                    bytes1(0x19),
```
Arachnid: As discussed the 0x19 and 0x00 bytes can be removed since we can't use this as a raw 191 type 0 signature.

contracts/reverseRegistrar/SignatureReverseResolver.sol
```diff
@@ -0,0 +1,276 @@
...
+        address addr,
+        uint256 inceptionDate,
+        bytes memory signature
+    ) {
+        isAuthorisedWithSignature(hash, addr, inceptionDate, signature);
+        _;
+    }
+
+    function isAuthorised(address addr) internal view virtual returns (bool) {
+        revert("This function needs to be overridden");
```
Arachnid: IIRC virtual methods don't need function bodies, which will cause a compile time error if you fail to override it.

contracts/reverseRegistrar/SignatureReverseResolver.sol
Arachnid: You're right; to be properly secure, inception dates would need to be recorded on each individual record (in the case of parameterised values such as text, on each key of each record); otherwise someone can replay just the last event, making it impossible to set earlier ones.

If someone omits a message, someone else can supply it - the issue is cases where omitting an earlier message but including a later one makes it impossible to supply the missing message later and arrive at the same end result.

Arachnid (changes requested):

contracts/reverseRegistrar/SignatureReverseResolver.sol
Arachnid: What do you mean?

Arachnid:
> I think it should be incorporated, there's no reason for it to be a separate parameter here. at https://github.com/ensdomains/ens-contracts/pull/335#discussion_r1543548879 , addr is used to derive node at bytes32 node = _getNamehash(addr); line 73 @Arachnid

Line 73 of what file? I can't see anywhere that would require this to be a separate parameter to the outer signature.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: To support EIP191 I prepended `0x19 <0x00> <intended validator address>`. Do you want me to revert all the changes, or do you still want me to keep `<intended validator address>` ?

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: As you can see from the code below, `addr` is used to derive `node = _getNamehash(addr);` hence it should be passed as part of `isAuthorisedWithSignature` rather than included in the `hash` (Of course we can technically include the address into hash as well as pass to the argument but seems redundant)
```
    function isAuthorisedWithSignature(
        bytes32 hash,
        address addr,
        uint256 inceptionDate,
        bytes memory signature
    ) internal view returns (bool) {
        bytes32 message = getSignedMessageHash(hash, addr, inceptionDate);
        bytes32 node = _getNamehash(addr);
```

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: `otherwise someone can replay just the last event, making it impossible to set earlier ones.` That would be problematic if we have a counter logic (eg: `n=n+1`) where the state depends on previous state. However, all `setter*` message simply overrides the previous state so there isn't much point supplying the missing message later.

I also didn't fully understand for ` inception dates would need to be recorded on each individual record`

Are you suggesting to change from

```
    function _setLastUpdated(bytes32 node, uint256 inceptionDate) internal {
        lastUpdated[node] = inceptionDate;
    }
```

to 

```
    function _setTextLastUpdated(bytes32 node, string key, uint256 inceptionDate) internal {
        lastUpdated[node][key] = inceptionDate;
    }
```

then change `isAuthorisedWithSignature` to take `key` ?

Arachnid: > That would be problematic if we have a counter logic (eg: n=n+1) where the state depends on previous state. However, all setter* message simply overrides the previous state so there isn't much point supplying the missing message later.

They don't, though - setting the address for one coin type invalidates all other previous messages to set addresses for other coin types. The same goes for text records.

Either the last updated value needs to be per-key as you illustrate, or we need to change this to function differently - for example, by taking a complete set of records to set in a single operation, and increment the version at the same time. If going with the former, probably the last updated check should be moved outside `isAuthorisedWithSignature`.

contracts/reverseRegistrar/SignatureReverseResolver.sol
Arachnid: Yes, keep the validator address, because we want the signature to not be replayable across different resolver contracts.

contracts/reverseRegistrar/L2ReverseRegistrar.sol
Arachnid: Renaming it seems like a good solution.

contracts/reverseRegistrar/SignatureReverseResolver.sol
makoto: Okay, it's getting a bit more complicating than simple fixes. You have been pointing out many changes to the functions not related to this PR but introduced at https://github.com/ensdomains/ens-contracts/pull/265 (you only noticed because I moved the code to different file).

May I suggest to leave it for the separate PR by @jefflau

Arachnid (approved):

PR #344, opened 2024-04-04
Use addressOfClone2 instead of predictAddress

Chomtana:
[DelegatableResolverFactory](https://github.com/ensdomains/ens-contracts/blob/delegatable-resolver-with-factory/contracts/resolvers/DelegatableResolverFactory.sol) contract is using `predictAddress` function that has been renamed to `addressOfClone2` on [upstream repository](https://github.com/wighawag/clones-with-immutable-args/blob/master/src/ClonesWithImmutableArgs.sol)


Arachnid (approved):
Thank you!

PR #352, opened 2024-05-22
feat: test improvements

TateB:
changes:
- switched from yarn to bun for package manager
- using typescript esm setup for hardhat (see https://github.com/NomicFoundation/hardhat/issues/3385#issuecomment-1841380253)
- test changes
  - all now in typescript
  - removed `ethereum-waffle`, replaced with `@ensdomains/hardhat-chai-matchers-viem` for fully typed test matchers
  - using deploy fixtures, instead of shared setup in `beforeEach`
  - refactored any tests that required the result of a prior test to run correctly, all tests are now entirely independent
  - removed usage of ethers (and in some cases web3) in favour of viem for better types
  - fixed any tests that were not running correctly, or were not testing what was expected
  - ci tests run in parallel
- deploy changes
  - using viem
  - `viem.deploy` function [1]
    - viem's `deployContract` for devnet
    - hardhat-deploy's `deploy` for live networks
      - for pending transaction save which can be resumed
  - using anvil for testing deploy on ci (significantly faster than hardhat, but doesn't have the same stack tracing)
  - multicall TLD setup for speed + gas efficiency
- removed unused dependencies (thanks @talentlessguy)

note [1]:
using `deploy` from hardhat-deploy on a devnet deploy is very slow, which i think is related to ethers provider polling parameters for a receipt (it becomes apparent past the `set_tlds` deploy script). this is why the usage is switched over for devnet to just use viem's `deployContract`, but still saves the deployment into hardhat-deploy's deployment manager which allows it to easily be accessed later. live network deploys still use `deploy`, but speed isn't relevant for them.


deploy/dnsregistrar/20_set_tlds.ts
```diff
@@ -1,61 +1,142 @@
...
+  namehash,
+  parseAbi,
+  parseEther,
+  type Address,
+  type Hash,
+  type Hex,
+} from 'viem'
+import { dnsEncodeName } from '../../test/fixtures/dnsEncodeName.js'
+
+const multicallAddress = '0xcA11bde05977b3631167028862bE2a173976CA11'
```
Arachnid, line 19: A comment stating where this comes from would be helpful.

Arachnid (approved):

PR #367, opened 2024-07-02
fix incompatible node version in .nvmrc

msgul:
Node version in .nvmrc (v14.15.0) is no longer compatible with the module.
Screenshot 2024-07-02 at 12 12 49
v16.20.2 seems to be working fine.


v1rtl:
we're no longer using yarn, we use bun now, so the issue should go away

PR #379, opened 2024-08-06
feature(crosschain resolver): simplify reverse resolver

TateB:
changes:
- `SignatureReverseResolver`
  - removed `setText*` functions
  - removed versionable records
    - no longer required without `setText`, since there is only one field set
  - removed `clearRecords` functions
    - no longer required without record versioning
  - optimisation: removed usage of `authorisedSignature` modifier and `isAuthorisedWithSignature` func, switched to inline
    - `isAuthorisedWithSignature` used in the `authorisedSignature` modifier required `node` for the address which was calculated within the function itself. `node` however was also used elsewhere in subsequent execution outside the modifier, and since modifiers can't return values it was unreachable. initially switched to just calling the `isAuthorisedWithSignature` function inline, but it's only used once now (from the removal of setText/clearRecords), so it's just inline.
  - optimisation: changed signature structure to remove the usage of the double-hashed function selector/name (i.e. `bytes32` id replaced with `bytes4, string`)
    - not really sure why it was like this in the first place?
  - optimisation: inceptionDate now in seconds (to align with block timestamp) instead of milliseconds
    - previously afaik ms was a requirement because of multicalls and subsequent calls within the same block, but now that's no longer really useful
  - optimisation: `name` variable as `calldata` instead of `memory`
- `L2ReverseResolver`
  - optimisation: removed use of `ownerAndAuthorisedWithSignature ` modifier function, logic now inline
  - optimisation: `name` variable as `calldata` instead of `memory`
  - optimisation: inceptionDate now in seconds
- `ReverseRegistrar`
  - removed `relayer` argument in `setNameForAddrWithSignature` and `claimForAddrWithSignature`
    - it was redundant since we check that the relayer is `msg.sender` anyway, and if `msg.sender` isn't the relayer variable the signed message won't be verified
  - optimisation: `name` variable as `calldata` instead of `memory`
- added ts tests
- added license to contracts
- moved `sha3HexAddress` func from `LowLevelCallUtils` lib to standalone `AddressUtils`
  - there was no overlap in usage of that function and the rest of the library, and the function alone isn't really a low level call utility func


Arachnid (approved):

adraffy (approved):

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+            if (
+                name.length != _dnsEncodedReverseNameLength + ADDRESS_LENGTH + 1
+            ) revert Unreachable(name);
+            if (keccak256(name[41:]) != _dnsEncodedReverseNameHash)
+                revert Unreachable(name);
+        }
+
+        if (selector == INameResolver.name.selector) {
+            if (isNamespaceCall) return abi.encode("");
+            (address addr, ) = HexUtils.hexToAddress(
```
Arachnid: This needs to capture the second return value, `value` ,and throw `Unreachable(name)` if it's not a valid hex string.

Arachnid (changes requested):

Arachnid (approved):
Approving so the current PR can be merged to reflect deployed code.

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+    ERC165,
+    Ownable
+{
+    using GatewayFetcher for GatewayRequest;
+
+    /// @notice The ENS registry contract.
+    ENS immutable ens;
+
+    /// @notice The gateway verifier contract, unique to each L2 chain.
+    IGatewayVerifier immutable verifier;
```
adraffy: This should be public

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+
+    /// @notice The target registrar contract on the L2 chain.
+    address immutable target;
```
adraffy: This should be public, and maybe `l2Registrar` or `registrarAddress`?

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+    string[] internal _urls;
+
+    /// @notice Emitted when the gateway URLs are changed.
+    event GatewayURLsChanged(string[] urls);
+
+    /// @notice Thrown when the name is not reachable in this resolver's namespace.
+    error Unreachable(bytes name);
+
+    /// @notice Thrown when the resolver profile is unknown.
+    error UnknownResolverProfile(bytes4 selector);
```
adraffy: Can we use `UnreachableName` and `UnsupportedResolverProfile` instead?

https://github.com/adraffy/ensips/blob/ensip/ensip-10-errors/ensips/x.md

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+            if (isNamespaceCall) return abi.encode("");
+            (address addr, ) = HexUtils.hexToAddress(
+                name,
+                1,
+                ADDRESS_LENGTH + 1
+            );
+            // Always throws, does not need to return.
+            _fetchName(addr);
+        } else if (selector == IAddressResolver.addr.selector) {
+            if (isNamespaceCall) return abi.encode(abi.encodePacked(target));
```
adraffy: Should we parse out `coinType` from the reverse name, and then only answer if the queried coin type matches?  Otherwise, this supplies an EVM address for all coin types.

contracts/reverseRegistrar/L1ReverseResolver.sol
```diff
@@ -0,0 +1,213 @@
...
+        return defaultReverseRegistrar.nameForAddr(addr);
+    }
+
+    /// @inheritdoc IERC165
+    function supportsInterface(
+        bytes4 interfaceId
+    ) public view override returns (bool) {
+        return
+            interfaceId == type(IExtendedResolver).interfaceId ||
+            super.supportsInterface(interfaceId);
```
adraffy: I think we should define an interface id that indicates that this is "recursive CCIP-read safe" so this contract can be called directly by the UR.

Maybe `0xe6894ef0`  ?  I'm not sure if there's a better derivation scheme for functionless interfaces...
* https://adraffy.github.io/keccak.js/test/demo.html#algo=evm&s=eip3668.wrappable&escape=1&encoding=utf8
* https://www.4byte.directory/signatures/?bytes4_signature=0xe6894ef0

contracts/reverseRegistrar/L1ReverseResolver.sol
TateB: if `isNamespaceCall` is true, it's querying with the reverse namespace of this resolver itself (i.e. `coinType` will always be correct)

meny3437-blip (commented):
future @meny3437-blip

PR #387, opened 2024-09-13
Initial revision of MigrationHelper

Arachnid:
MigrationHelper would allow people to opt in to the ENSv2 migration ahead of time by calling `setApprovalForAll` on the .eth registrar or Name Wrapper targeting this contract. This contract will be owned by the DAO; at migration time, the DAO can specify a migration target address (likely the `L1MigrationController`) and a controller contract or address that is authorized to make calls to migrate domains to this contract.

Notably, this introduces a trusted dependency - a malicious DAO could use this contract to steal names from users who have elected to use it.


contracts/utils/MigrationHelper.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function migrateNames(
+        address owner,
+        uint256[] memory tokenIds,
+        bytes memory data
+    ) external onlyController {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
+            registrar.safeTransferFrom(
+                owner,
```
makoto: Is the `owner` the owner of this contract (aka DAO) or the actual name owner? If the latter, you may want to change to `nameowner` or something as `onlyOwner` at `setMigrationTarget` and `owner` here have different semantics.

jefflau: Looks like it's the actual name owner, because the owner of the name is what you need to transfer it

jefflau (approved):
LGTM. I guess we need E2E tests of this being approved and then a dummy migration controller to make sure everything works as expected

contracts/utils/MigrationHelper.sol
Arachnid: `owner` in this context is the owner of the name, yes. It's passed in as a function parameter.

Arachnid:
@jefflau @TateB PTAL

TateB (changes requested):
test needs to be in typescript

TateB (approved):

PR #417, opened 2025-03-12
IUniversalResolver and modified UniversalResolver

adraffy:
- add [IUniversalResolver](https://github.com/ensdomains/ens-contracts/blob/bet-193/universal-resolver-interface/contracts/universalResolver/IUniversalResolver.sol)
- move existing `UniversalResolver` to [contracts/universalResolver](https://github.com/ensdomains/ens-contracts/tree/bet-193/universal-resolver-interface/contracts/universalResolver)
- `UniversalResolver` is `IUniversalResolver`
    - `ResolverNotFound()` &rarr; `ResolverNotFound(bytes name)` 
    - `ResolverNotContract()` &rarr; `ResolverNotContract(bytes name, address resolver)`
    - removed `ResolverWildcardNotSupported`, using `ResolverNotFound` instead
    - single resolve uses `HttpError(uint16,string)` if `HttpError(<1 item>)`
    - `reverse(bytes reverseName)` &rarr; `reverse(bytes lookupAddress, uint256 coinType)`
    - `reverse(bytes reverseName, string[])` &rarr; `reverse(bytes lookupAddress, uint256 coinType, string[])`
    - changed `reverse()` return type to match `IUniversalResolver`
    - changed `reverse()` behavior to revert `ReverseAddressMismatch()`
    - added revert `Error("reverse: expected coinType 60")`
    - added revert `Error("reverse: expected EVM address")`


TateB (approved):

contracts/universalResolver/IUniversalResolver.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
+
+/// @notice Interface for the UniversalResolver.
+interface IUniversalResolver {
+    /// @notice A resolver could not be found for the supplied name.
+    error ResolverNotFound(bytes name);
+
+    /// @notice The resolver is not a contract.
+    error ResolverNotContract(bytes name, address resolver);
```
Arachnid, line 10: Does this also cover the case where the resolver doesn't implement the required interface per ERC-165?

contracts/universalResolver/IUniversalResolver.sol
```diff
@@ -0,0 +1,52 @@
...
+    /// @notice A HTTP error occurred on the batch gateway.
+    error HttpError(uint16 status, string message);
+
+    /// @notice Performs ENS name resolution for the supplied name and resolution data.
+    /// @param name The name to resolve, in normalised and DNS-encoded form.
+    /// @param data The resolution data, as specified in ENSIP-10.
+    ///             For a multicall, the data should be encoded as `(bytes[])`.
+    /// @return result The result of the resolution.
+    ///                For a multicall, the result is encoded as `(bytes[])`.
+    /// @return resolver The resolver that was used to resolve the name.
```
Arachnid: This should note that callers need to support CCIP-Read.

contracts/universalResolver/IUniversalResolver.sol
```diff
@@ -0,0 +1,52 @@
...
+    ) external view returns (bytes memory result, address resolver);
+
+    /// @notice Performs ENS reverse resolution for the supplied address and coin type.
+    /// @param lookupAddress The address to reverse resolve, in encoded form.
+    /// @param coinType The coin type to use for the reverse resolution.
+    ///                 For ETH, this is 60.
+    ///                 For other EVM chains, coinType is calculated as `0x80000000 | chainId`.
+    /// @return primary The reverse resolution result.
+    /// @return resolver The resolver that was used to resolve the name.
+    /// @return reverseResolver The resolver that was used to resolve the reverse name.
```
Arachnid: This should note that callers need to support CCIP-Read.

contracts/universalResolver/UniversalResolver.sol
Arachnid, line 1: Can you make `UniversalResolver` implement `IUniversalResolver`?

Arachnid (changes requested):

contracts/universalResolver/IUniversalResolver.sol
adraffy: in [UR2](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/contracts/universalResolver/ForwardResolution.sol#L51-L64): 
* I use [OZ ERC165Checker](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.1/contracts/utils/introspection/ERC165Checker.sol#L34) first to check extended
     * which does 3 checks: 165, !invalid, IExtended
* I use `resolver.code.length` to check if contract

so if it "doesn't implement the required interface per ERC-165", then it's not extended, and:
* if it's not the names resolver &rarr; `ResolverNotFound()`
* if it has no code &rarr; `ResolverNotContract()`

in UR1, logic is unchanged except that `ResolverWildcardNotSupported` is removed since it's not in the interface (better covered by `ResolverNotFound` imo)

contracts/universalResolver/UniversalResolver.sol
adraffy: okay, you'll have to tell me if that's too much collateral damage to the UR1

contracts/universalResolver/IUniversalResolver.sol
Arachnid: What if it's not extended, is the name's resolver, is a contract, but just doesn't implement that function?

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -204,74 +203,79 @@ contract UniversalResolver is ERC165, Ownable {
...
-        bytes calldata reverseName,
+        bytes memory lookupAddress,
+        uint256 coinType,
         string[] memory gateways
-    ) public view returns (string memory, address, address, address) {
-        bytes memory encodedCall = abi.encodeCall(
-            INameResolver.name,
-            reverseName.namehash(0)
+    ) public view returns (string memory, address, address) {
+        require(coinType == 60, "reverse: expected coinType 60");
```
Arachnid, line 222: Can you open an issue in Linear to track adding support for reverse lookups of other cointypes to the universal resolver?

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -583,21 +589,28 @@ contract UniversalResolver is ERC165, Ownable {
...
     }
 
     function _checkResolveSingle(Result memory result) internal pure {
         if (!result.success) {
-            if (bytes4(result.returnData) == HttpError.selector) {
-                bytes memory returnData = result.returnData;
-                assembly {
-                    revert(add(returnData, 32), mload(returnData))
+            bytes memory v = result.returnData;
+            if (bytes4(v) == 0xca7a4e75) {
```
Arachnid, line 601: You should be able to use `HttpError.selector` in place of the fixed value.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -583,21 +589,28 @@ contract UniversalResolver is ERC165, Ownable {
...
+                HttpErrorItem[] memory items = abi.decode(
+                    v.substring(4, v.length - 4),
+                    (HttpErrorItem[])
+                );
+                if (items.length == 1) {
+                    revert HttpError(
+                        items[0].status,
+                        items[0].message
+                    );
                 }
```
Arachnid, line 611: What's this in aid of? Why not revert with the original array?

Arachnid (changes requested):

contracts/universalResolver/UniversalResolver.sol
adraffy: I think the intended error is `HttpError(status, message)`, the `HttpError(tuple(status, message)[])` is a UR1 creation.

contracts/universalResolver/UniversalResolver.sol
adraffy: There are 2 `HttpError`'s.  There appears to be no way to select that outer error above.

contracts/universalResolver/UniversalResolver.sol
Arachnid: What happens if `items.length != 1`?

contracts/universalResolver/UniversalResolver.sol
Arachnid: Is it necessary to have two?

PR #420, opened 2025-03-16
bug: `BytesUtils.compare()` Arithmetic Overflow

adraffy:
The difference `int256 - int256` exceeds `int256` range.

Example: `type(int256).min - 1 < type(int256).min`

---

* change `BytesUtils.compare()` to fix overflow
* add test case for maximum signed difference (caused error before)
* to test: `bun run test test/dnssec-oracle/TestSolidityTests.ts`

### DIFF
--- a/contracts/utils/BytesUtils.sol
+++ b/contracts/utils/BytesUtils.sol
@@ -77,11 +77,11 @@ library BytesUtils {
         return compare(self, 0, self.length, other, 0, other.length);
     }
 
-    /**
-     * @dev +1 if `other` comes lexicographically after `self`,
-     *      -1 if `other` comes before,
-     *      or 0 if the contents of the two bytes are equal.
-     *      Comparison is done bytewise.
+    /*
+     * @dev Returns a positive number if `other` comes lexicographically after
+     *      `self`, a negative number if it comes before, or zero if the
+     *      contents of the two bytes are equal. Comparison is done per-rune,
+     *      on unicode codepoints.
      * @param self The first bytes to compare.
      * @param offset The offset of self.
      * @param len    The length of self.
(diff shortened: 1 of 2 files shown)


contracts/utils/BytesUtils.sol
```diff
@@ -77,11 +77,11 @@ library BytesUtils {
...
-    /*
-     * @dev Returns a positive number if `other` comes lexicographically after
-     *      `self`, a negative number if it comes before, or zero if the
-     *      contents of the two bytes are equal. Comparison is done per-rune,
-     *      on unicode codepoints.
+    /**
+     * @dev +1 if `other` comes lexicographically after `self`,
+     *      -1 if `other` comes before,
+     *      or 0 if the contents of the two bytes are equal.
+     *      Comparison is done bytewise.
```
Arachnid: This should stay as it was - we don't want to unduly constrain implementations.

contracts/utils/BytesUtils.sol
```diff
@@ -123,21 +123,25 @@ library BytesUtils {
                 b := mload(otherptr)
             }
             if (a != b) {
-                // Mask out irrelevant bytes and check again
-                uint256 mask;
-                if (shortest - idx >= 32) {
-                    mask = type(uint256).max;
-                } else {
-                    mask = ~(2 ** (8 * (idx + 32 - shortest)) - 1);
+                uint256 rest = shortest - idx;
```
Arachnid, line 126: Can you leave the comment intact?

contracts/utils/BytesUtils.sol
```diff
@@ -123,21 +123,25 @@ library BytesUtils {
...
+                if (rest < 32) {
+                    rest = (32 - rest) << 3; // bits to drop
+                    a >>= rest; // drop them
+                    b >>= rest;
                 }
-                int256 diff = int256(a & mask) - int256(b & mask);
-                if (diff != 0) return diff;
+                if (a < b) return -1;
+                if (a > b) return 1;
```
Arachnid: Please put all if clauses in braces.

contracts/utils/BytesUtils.sol
```diff
@@ -123,21 +123,25 @@ library BytesUtils {
...
             otherptr += 32;
         }
 
-        return int256(len) - int256(otherlen);
+        return
+            len == otherlen
+                ? int256(0)
+                : len < otherlen
+                    ? int256(-1)
+                    : int256(1);
```
Arachnid: If we don't change the docstring, this change isn't needed either.

Arachnid (changes requested):

Arachnid (approved):

PR #421, opened 2025-03-17
UniversalResolver w/CCIPBatcher

adraffy:
- add [IUniversalResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/universalResolver/IUniversalResolver.sol)
- replace [UniversalResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/universalResolver/UniversalResolver.sol) and [TestUniversalResolver.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/universalResolver/TestUniversalResolver.ts) with new versions
- add mainnet tests [TestUniversalResolver.remote.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/universalResolver/TestUniversalResolver.remote.ts) + [mainnet.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/universalResolver/mainnet.ts)
     - use `bun run test:remote`

---

- Components
    - add [CCIPReader.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/ccipRead/CCIPReader.sol) to simplify recursive EIP-3668
    - add [CCIPBatcher.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/ccipRead/CCIPBatcher.sol) to simplify batch gateway
- Libraries
    - modified [HexUtils.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/utils/HexUtils.sol) and [TestHexUtils.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/utils/TestHexUtils.ts) with more hex functions
    - add [NameCoder.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/utils/NameCoder.sol) and [TestNameCoder.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/utils/TestNameCoder.ts) think`BytesUtils` and `HexUtils` w/encrypted labels
    - add [ENSIP19.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/utils/ENSIP19.sol) and [TestENSIP19.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/utils/TestENSIP19.ts) for reverse names
    - add [EIP3668.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/ccipRead/EIP3668.sol) for `OffchainLookup` decoding

---

* add [IBatchGateway.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/ccipRead/IBatchGateway.sol)
* add [localBatchGateway.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/fixtures/localBatchGateway.ts) and [TestLocalBatchGateway.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/test/utils/TestLocalBatchGateway.ts) for gatewayless testing
* add [DummyShapeshifterResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-318-standard-ur-alt/contracts/universalResolver/mocks/DummyShapeshiftResolver.sol) to emulate many resolver types

---

* change `package.json` to explicit bun use, add  `"patchedDependencies"`
* replace `getReverseName.ts` fixture with `ensip19.ts` fixture


contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+import {CCIPReader, EIP3668, OffchainLookup} from "../ccipRead/CCIPReader.sol";
+
+contract Batchcall is CCIPReader, IBatchcall, IERC165 {
+    function supportsInterface(bytes4 x) external pure returns (bool) {
+        return
+            type(IERC165).interfaceId == x || type(IBatchcall).interfaceId == x;
+    }
+
+    function batch(
+        Thread[] memory threads,
```
Arachnid: `Thread` is a bit of an overloaded term, can we use `Call` or something similar?

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+        string[] memory gateways
+    ) external view returns (Thread[] memory) {
+        for (uint256 i; i < threads.length; i++) {
+            Thread memory t = threads[i];
+            (bool ok, bytes memory v) = t.target.staticcall(t.call);
+            if (v.length == 0) {
```
Arachnid: Shouldn't this distinguish `ok` and `!ok` cases?

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+                v = abi.encodePacked(bytes4(t.call)); // remember the function
+                t.bits |= ThreadBits.DONE | ThreadBits.EMPTY_RESPONSE;
+            } else if (ok) {
+                t.bits |= ThreadBits.DONE;
+            } else if (bytes4(v) != OffchainLookup.selector) {
+                t.bits |= ThreadBits.DONE | ThreadBits.CALL_ERROR;
+            } else {
+                t.bits |= ThreadBits.OFFCHAIN;
+            }
```
Arachnid: It's a little clearer if these clauses are the other way around:

```suggestion
            } else if (bytes4(v) == OffchainLookup.selector) {
                t.bits |= ThreadBits.OFFCHAIN;
            } else {
                t.bits |= ThreadBits.DONE | ThreadBits.CALL_ERROR;
            }
```

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+    function batch(
+        Thread[] memory threads,
+        string[] memory gateways
+    ) external view returns (Thread[] memory) {
+        for (uint256 i; i < threads.length; i++) {
+            Thread memory t = threads[i];
+            (bool ok, bytes memory v) = t.target.staticcall(t.call);
+            if (v.length == 0) {
+                v = abi.encodePacked(bytes4(t.call)); // remember the function
+                t.bits |= ThreadBits.DONE | ThreadBits.EMPTY_RESPONSE;
```
Arachnid: The standard terminology here would be 'flags' rather than 'bits'.

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+                    p.sender,
+                    p.urls,
+                    p.callData
+                );
+            }
+        }
+        if (count > 0) {
+            assembly {
+                mstore(requests, count)
+            }
```
Arachnid:
```suggestion
            // Truncate `requests` to the number of elements actually used
            assembly {
                mstore(requests, count)
            }
```

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+                gateways,
+                abi.encodeCall(IBatchGateway.query, (requests)),
+                this.batchCallback.selector,
+                abi.encode(threads, gateways)
+            );
+        }
+    }
+
+    function batchCallback(
+        bytes calldata ccip,
```
Arachnid: This should be called `response`.

contracts/batchGateway/Batchcall.sol
```diff
@@ -0,0 +1,116 @@
...
+                        )
+                    );
+                    if (v.length == 0) {
+                        v = abi.encodePacked(p.callbackFunction);
+                        t.bits |= ThreadBits.DONE | ThreadBits.EMPTY_RESPONSE;
+                    } else if (ok) {
+                        t.bits |= ThreadBits.DONE;
+                    } else if (bytes4(v) != OffchainLookup.selector) {
+                        t.bits |= ThreadBits.DONE | ThreadBits.CALL_ERROR;
+                    }
```
Arachnid: You should add a comment about why it's okay there's no `else` here to handle recursive calls.

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @author Modified from Unruggable https://github.com/unruggable-labs/CCIPReader.sol/
```
Arachnid: Can you change this to link to a specific blob?

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,133 @@
...
+
+/*
+MIT License
+
+Copyright (c) 2025 Unruggable
```
Arachnid:
```suggestion
Portions Copyright (c) 2025 Unruggable
Portions Copyright (c) 2025 ENS Labs Ltd
```

I don't think we need to include the full license text - just these copyright declarations and the identification of the license.

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,133 @@
...
+/// 2. call `ccipRead()` similar to `staticcall()`
+/// 3. implement all response logic in callback
+/// 4. ensure that return type of calling function == callback function
+/// 5. any return value of `ccipRead()` is already abi-encoded
+
+/// Use the following code to return it directly:
+/// ```solidity
+/// bytes memory v = ccipRead(...);
+/// assembly { return(add(v, 32), mload(v)) }
+/// ```
```
Arachnid: I think this should be part of the docstring for `CCIPReader`.

Also, I think you can probably implement the 'return directly' code with a library function.

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,133 @@
...
+            );
+        }
+        // OR the call to the 'real' target reverts with a different error selector
+        // OR the call to OUR callback reverts with ANY error selector
+        if (!ok) {
+            assembly {
+                revert(add(v, 32), mload(v))
+            }
+        }
+    }
```
Arachnid:
```suggestion
        // If the call to the 'real' target reverts with a different error selector
        // OR the call to OUR callback reverts with ANY error selector
        if (!ok) {
            assembly {
                revert(add(v, 32), mload(v))
            }
        }

        // IF we have gotten here, the 'real' target does not revert with an `OffchainLookup` error
        // The exit point of this architecture is  OUR callback in the 'real'
        // We pass through the response to that callback
        (ok, v) = address(this).staticcall(
            abi.encodeWithSelector(myCallbackFunction, v, myExtraData)
        );
    }
```

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// for internal use
+struct Context {
+    address target;
+    bytes4 callbackFunction;
+    bytes extraData;
+    bytes4 myCallbackFunction;
+    bytes myExtraData;
+}
```
Arachnid: This should go inside the contract, or be above the contract if that's not possible.

contracts/ccipRead/EIP3668.sol
```diff
@@ -0,0 +1,56 @@
...
+
+/// Example usage:
+/// ```solidity
+/// bytes memory v = hex"...";
+/// | if (bytes4(v) == OffchainLookup.selector) {
+/// |    EIP3668.Params memory params = EIP3668.decodeWithSelector(v);
+/// | }
+/// ```
+
+library EIP3668 {
```
Arachnid, line 16: Does this need to be a separate library to `CCIPReader`?

contracts/ccipRead/EIP3668.sol
```diff
@@ -0,0 +1,56 @@
...
+
+function dropLeadingBytes(
+    bytes memory data,
+    uint256 size
+) pure returns (bytes memory dropped) {
+    require(data.length >= size);
+    dropped = abi.encodePacked(data); // make a copy
+    assembly {
+        mstore(add(dropped, size), sub(mload(dropped), size)) // subtract length
+        dropped := add(dropped, size) // add offset
```
Arachnid: Consider using `memcpy` from `ens-contracts/utils/BytesUtils.sol` instead.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+        batchcall = _batchcall;
+        registry = ens;
+        batchGateways = gateways;
+    }
+
+    function supportsInterface(bytes4 x) external pure returns (bool) {
+        return
+            type(IERC165).interfaceId == x ||
+            type(IUniversalResolver).interfaceId == x;
+    }
```
Arachnid: This needs to inherit `ERC165` and include a call to `super`.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+
+    function setBatchGateways(string[] memory gateways) external onlyOwner {
+        batchGateways = gateways;
+    }
+
+    function findResolver(
```
Arachnid, line 45: All public functions need docstrings. If they're documented in the interface it can just have a reference to the interface.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+    {
+        Lookup memory lookup = lookupResolver(name);
+        resolver = lookup.resolver;
+        namehash = lookup.node;
+        finalOffset = lookup.offset;
+    }
+
+    struct Lookup {
+        bytes name; // dns-encoded name (safe to decode)
+        uint256 offset; // byte offset into name for basename
```
Arachnid: Can you clarify what `basename` is?

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+        bytes32 node = NameCoder.namehash(name, 0);
+        lookup.name = name;
+        lookup.node = node;
+        while (true) {
+            resolver = registry.resolver(node);
+            if (resolver != address(0)) break; // found a resolver
+            uint256 size = uint8(name[offset]);
+            if (size == 0) revert ResolverNotFound(name); // no match
+            offset += 1 + size;
+            node = NameCoder.namehash(name, offset);
```
Arachnid: This process is O(n^2). Can it be restructured as a recursive function?

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+                resolver,
+                type(IExtendedResolver).interfaceId
+            )
+        ) {
+            lookup.extended = true;
+        } else if (offset != 0) {
+            revert ResolverNotFound(name); // non-extended resolver requires exact match
+        } else if (resolver.code.length == 0) {
+            revert ResolverNotContract(name, resolver);
+        }
```
Arachnid: I wonder if this check should be done in the caller, rather than here?

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+        assembly {
+            return(add(v, 32), mload(v))
+        }
+    }
+
+    struct ReverseArgs {
+        bytes encodedAddress;
+        uint256 coinType;
+        string[] gateways;
+    }
```
Arachnid: Structs should be above functions.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+        bytes memory v = _requireResponse(threads[0]);
+        bytes memory primaryAddress;
+        if (bytes4(threads[0].call) == IAddrResolver.addr.selector) {
+            address addr = abi.decode(v, (address));
+            if (addr != address(0)) {
+                primaryAddress = abi.encodePacked(addr);
+            }
+        } else {
+            primaryAddress = abi.decode(v, (bytes));
+        }
```
Arachnid: I assume this is assuming it must be `IAddressResolver.addr.selector`, but we should explicitly check.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,349 @@
...
+        if (keccak256(reverseAddress) != keccak256(primaryAddress)) {
```
Arachnid: `BytesUtils` has an `equals` method for this.

contracts/utils/HexUtils.sol
```diff
@@ -98,27 +98,94 @@ library HexUtils {
...
+    /**
+     * @dev Converts an uint256 to a hex string without zero padding
+     *      eg. unpaddedUintToHex(0, true)  = "0"
+     *      eg. unpaddedUintToHex(1, true)  = "1"
+     *      eg. unpaddedUintToHex(0, false) = "00"
+     *      eg. unpaddedUintToHex(1, false) = "01"
+     */
+    function unpaddedUintToHex(
+        uint256 value,
+        bool dropZeroNibble
```
Arachnid, line 122: nit: It's `nybble` not `nibble`.

contracts/utils/HexUtils.sol
```diff
@@ -98,27 +98,94 @@ library HexUtils {
...
+    ) internal pure returns (string memory hexString) {
+        uint256 temp = value;
+        uint256 shift;
+        for (uint256 b = 128; b >= 8; b >>= 1) {
+            if (temp < (1 << b)) {
+                shift += b; // number of zero upper bits
+            } else {
+                temp >>= b; // shift away lower half
+            }
+        }
```
Arachnid, line 132: This is very clever! Did you come up with it yourself?

Arachnid, line 132: This is very clever! Did you invent it yourself?

contracts/utils/NameCoder.sol
```diff
@@ -0,0 +1,175 @@
...
+/// @dev Library for encoding/decoding names
+
+/// An ENS name is stop-separated labels: eg. "aaa.bb.c"
+
+/// A DNS-encoded name is byte-length prefixed labels with a terminator byte
+/// eg. "\x03aaa\x02bb\x01c\x00"
+/// * maximum label length is 255 bytes
+/// * length = 0 is reserved for the terminator (root)
+
+/// To encode a label larger than 255 bytes, use a hashed label
```
Arachnid: We should make it clear that this is also a valid notation for shorter labels.

contracts/utils/NameCoder.sol
```diff
@@ -0,0 +1,175 @@
...
+///  w/ hashed labels: dns.length == 2 + ens.split('.').map(x => x.utf8Length).sum(n => n > 255 ? 66 : n)
+
+library NameCoder {
+    /// @dev The DNS-encoded name is malformed
+    error DNSDecodingFailed(bytes dns);
+
+    /// @dev Some label of the ENS name has an invalid size
+    error DNSEncodingFailed(string ens);
+
+    /// @dev Same as BytesUtils.readLabel() but supports hashed labels w/custom error
```
Arachnid: Can we update existing users of `readLabel` to use this, if it's an upgrade. Fine to add this as a backlog item for a separate PR.

contracts/utils/NameCoder.sol
```diff
@@ -0,0 +1,175 @@
...
+            if (start == 0) revert DNSEncodingFailed(ens);
+            assembly {
+                mstore8(start, 0) // terminal byte
+                mstore(dns, sub(start, add(dns, 31))) // truncate length
+            }
+        }
+    }
+
+    /// @dev returns 0 if error (handled by caller)
+    function _encodeLabel(
```
Arachnid: Perhaps something like `_createHashedLabel` would be clearer?

test/utils/TestHexUtils.ts
```diff
@@ -198,4 +198,48 @@ describe('HexUtils', () => {
...
+        const { hexUtils } = await loadFixture(fixture)
+        await expect(hexUtils.read.bytesToHex([data])).resolves.toStrictEqual(
+          data.slice(2),
+        )
+      })
+    }
+  })
+
+  describe('unpaddedUintToHex()', async () => {
+    for (const i of [0n, 1n, (1n << 256n) - 1n]) {
```
Arachnid: A few other even- and odd-length tests would be good here, in case there's a fencepost error that doesn't affect `0` and `1`.

Arachnid (changes requested):
Fantastic work here.

contracts/batchGateway/Batchcall.sol
adraffy: The case where a resolver does not implement a function results in `(true, 0x)`.

Rather than letting this blow up somewhere else because of a `abi.decode()` failure, it seems better to just turn it into an error so it can be traced.  I record the selector of the call (or the selector of the callback) and mark the error type so it can be handled later.

I'm not sure if `(false, 0x)` can happen?  Oh is that `revert(0, 0)`, hum.

contracts/utils/HexUtils.sol
adraffy: ya

contracts/utils/NameCoder.sol
adraffy: I was worried about making this change, as it's (2) things:  change to custom error and label hashing support.
We can combine these two libraries in a future PR.

contracts/batchGateway/Batchcall.sol
adraffy: Changed to flags

contracts/batchGateway/Batchcall.sol
adraffy: I changed to `Lookup`, short for `OffchainLookup`

contracts/batchGateway/Batchcall.sol
adraffy: I see your point but I kept the ordering, as it matches the ordering in the callback.  I added an explicit else-branch with a comment where the OFFCHAIN flag is already set but another `OffchainLookup` is requested.

* [ccipBatch()](https://github.com/ensdomains/ens-contracts/blob/84e5526f4069aad5da3030a067c669c3153da033/contracts/ccipRead/CCIPBatcher.sol#L38-L47) (entry point)
* [ccipBatchCallback()](https://github.com/ensdomains/ens-contracts/blob/84e5526f4069aad5da3030a067c669c3153da033/contracts/ccipRead/CCIPBatcher.sol#L113-L122) (callback loop)

contracts/batchGateway/Batchcall.sol
adraffy: Changed everywhere

contracts/ccipRead/EIP3668.sol
adraffy: It's actually pretty useful because there's no `abi.decodeError()` and the error as-is naturally destructures to 5 separate args which frequently causes "stack too deep" errors.

contracts/ccipRead/EIP3668.sol
adraffy: This is because I want `CCIPReader` to be an independent library.

contracts/universalResolver/UniversalResolver.sol
adraffy: I refer to the basename as the name corresponding to the resolver.  I updated the comment.

contracts/universalResolver/UniversalResolver.sol
adraffy: Refactored

adraffy:
This new refactor is **much cleaner** and replaces the external `Batchcall` with a `CCIPBatcher` helper so just 1 contract.

[`_unwrapResolve()`](https://github.com/ensdomains/ens-contracts/blob/84e5526f4069aad5da3030a067c669c3153da033/contracts/universalResolver/UniversalResolver.sol#L335-L345) is sketchy.  The assembly can be removed if I pass the unwrapped calldata in `extraData` rather than trying to decode the wrapped calldata.  However, I assume it's safe since the encoding is generated by the EVM not a client.

The [multicall result encoding](https://github.com/ensdomains/ens-contracts/blob/f9d268d9adc43dd8900af80593a8aceccaaaad8e/contracts/universalResolver/UniversalResolver.sol#L149-L151) may need some extra error processing.  An empty result will become a 4-byte response, but will be a function selector, not an error selector.

contracts/ccipRead/CCIPReader.sol
adraffy: This ordering is particular as both the initial call or the callback can fail but there's only one revert.

contracts/batchGateway/Batchcall.sol
adraffy: I [changed this logic](https://github.com/ensdomains/ens-contracts/blob/2af9e2e63b73434d6ca8a24669c24f6f77e6ebf3/contracts/ccipRead/CCIPBatcher.sol#L42-L47) and added [more tests](https://github.com/ensdomains/ens-contracts/blob/2af9e2e63b73434d6ca8a24669c24f6f77e6ebf3/test/universalResolver/TestUniversalResolver.ts#L162-L187).  
* empty return &rarr; `UnsupportedResolverProfile(bytes4)`
* empty revert &rarr; `ResolverError(0x)`

contracts/utils/NameCoder.sol
Arachnid: Can you add an item in Linear for this? I don't want us to forget and keep using two subtly different implementations indefinitely.

contracts/ccipRead/EIP3668.sol
Arachnid: Right, I was asking if `EIP3668` and `CCIPReader` could be merged into the same library, though.

contracts/ccipRead/EIP3668.sol
Arachnid: Why?

contracts/universalResolver/UniversalResolver.sol
Arachnid: They need *complete* docstrings. Please use `/** ... */` syntax and check with a linter that they're complete.

contracts/batchGateway/Batchcall.sol
Arachnid: I'm not sure what you mean about the ordering in the callback? Regardless it seems quite odd to have the != check in the 'if' and the == check in the `else`.

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,355 @@
...
+                info.resolver,
+                type(IExtendedResolver).interfaceId
+            )
+        ) {
+            info.extended = true;
+        } else if (info.offset != 0) {
+            revert ResolverNotFound(name); // immediate resolver requires exact match
+        } else if (info.resolver.code.length == 0) {
+            revert ResolverNotContract(name, info.resolver);
+        }
```
Arachnid, line 108: I think it may make sense for this function to not do this check, and leave it to the caller. Otherwise, life becomes more difficult for tools like the manager app that want to show information even if there's something wrong with it (like it's pointed at the wrong address).

Arachnid (changes requested):

contracts/utils/NameCoder.sol
adraffy: BET-333

contracts/universalResolver/UniversalResolver.sol
adraffy: Doesn't this function fully encapsulate all of the information?  Either `ResolverNotFound` or `ResolverNotContract` or it's valid.  This function is invoked twice during the `reverse()` flow so it seems like the right place.

`ResolverNotContract` doesn't reveal the offset but it could use the actual name

I added `name` to both of these errors because of the inner error when `requireResolver(primary)` fails.

contracts/batchGateway/Batchcall.sol
adraffy:
```js
if (is it done?) {
    // it's done
} else if (another way it can be done?) {
    // it's done
} else {
    // it's not done
}
```

contracts/universalResolver/UniversalResolver.sol
adraffy: I guess I made a mistake leaving a "single comment" as Github won't let me directly comment above.

* I'll change `EIP3668` to use `BytesUtils`.  My intention was to keep `CCIPReader` independent since it's a useful construction.  

* The `if-else` thing is the same as my comment above about handling all of the done cases first.

contracts/batchGateway/Batchcall.sol
adraffy: I swapped it

contracts/universalResolver/UniversalResolver.sol
adraffy: * I moved `decodeWithSelector` to `CCIPReader` which keeps `EIP3668` dependency-free.
* I changed the `if-else` orderings
* I added a ton of docstring annotations.
* I addressed the [multicall result encoding](https://github.com/ensdomains/ens-contracts/blob/186cf1b33df150544b99292d81a062480344823c/contracts/universalResolver/UniversalResolver.sol#L160-L162) concern above.
* I simplified the `ccipRead()` return logic.

contracts/ccipRead/CCIPReader.sol
adraffy: I was thinking there might be some reason to have control over it but you're right, its way cleaner if `CCIPReader` just returns.

contracts/ccipRead/EIP3668.sol
adraffy: EIP3668 is primarily just an `OffchainLookup` header w/ struct decoder now.  Other logic moved to `CCIPReader`.

contracts/ccipRead/EIP3668.sol
adraffy: Removed, see below.

contracts/ccipRead/EIP3668.sol
Arachnid: But why do they need to be separate libraries at all? Can't one library contain both pieces of functionality? And why are you determined to keep this file dependency free?

Arachnid (changes requested):
Can you please change the comment style for doccomments from `///` to `/** ... */`? This is the convention used elsewhere in the codebase.

contracts/universalResolver/UniversalResolver.sol
Arachnid: The return value contains more information than the error - the offset for the resolver found, at a minimum. It's also easier for interfaces such as the manager app to do this check themselves rather than process two entirely different response formats. If we're calling the resolver, the Solidity compiler also redundantly does this check before making a call. I'd personally be more comfortable leaving it out.

Arachnid (changes requested):

contracts/universalResolver/UniversalResolver.sol
adraffy: I think `findResolver()` is sufficient for the use-case where you just want the resolver address and offset.  This is the same as the current UR and tate's rewrite. 

`requireResolver()` was an internal function I made visible because it seemed more useful w/o breaking the current behavior.

contracts/ccipRead/EIP3668.sol
adraffy: Not every use of `OffchainLookup` needs `CCIPReader`.

`OffchainLookup` should have a header file instead of being copy-pasted everywhere.

contracts/universalResolver/UniversalResolver.sol
Arachnid: Fair enough.

Arachnid (approved):

contracts/ccipRead/CCIPBatcher.sol
```diff
@@ -0,0 +1,159 @@
...
+                            lu.flags |= FLAG_DONE;
+                            if (v.length == 0) {
+                                v = abi.encodePacked(p.callbackFunction);
+                                lu.flags |= FLAG_EMPTY_RESPONSE;
+                            }
+                        } else if (bytes4(v) == OffchainLookup.selector) {
+                            // another offchain request
+                        } else {
+                            lu.flags |= FLAG_DONE | FLAG_CALL_ERROR;
+                        }
```
TateB: is this pattern just to follow `ccipBatch`? why not
```
} else if (bytes4(v) != OffchainLookup.selector) {
    lu.flags |= FLAG_DONE | FLAG_CALL_ERROR;
}
```

contracts/ccipRead/CCIPBatcher.sol
```diff
@@ -0,0 +1,159 @@
...
+     * @return batch The batch where every lookup is "done".
+     */
+    function ccipBatchCallback(
+        bytes calldata response,
+        bytes calldata extraData
+    ) external view returns (Batch memory batch) {
+        (bool[] memory failures, bytes[] memory responses) = abi.decode(
+            response,
+            (bool[], bytes[])
+        );
```
TateB, line 116: is this for backwards compatibility with the existing batch gateway? it's not really necessary, can just test against `responses[i].length` instead

contracts/ccipRead/CCIPReader.sol
```diff
@@ -0,0 +1,143 @@
...
+     * @dev Special-purpose value for identity callback: `f(x) = x`.
+     */
+    bytes4 constant IDENTITY_FUNCTION = bytes4(0);
+
+    /**
+     * @dev Same as `ccipRead()` but the callback function is the identity.
+     */
+    function ccipRead(address target, bytes memory call) internal view {
+        ccipRead(target, call, IDENTITY_FUNCTION, "");
+    }
```
TateB: maybe this makes more sense as `call`/etc rather than`ccipRead`, since it's just a ccip-read enabled call func

contracts/universalResolver/UniversalResolver.sol
```diff
@@ -0,0 +1,421 @@
...
+                    (info.name, calls[i])
+                )
+                : calls[i];
+        }
+        ccipRead(
+            address(this),
+            abi.encodeCall(this.ccipBatch, (batch)),
+            this.resolveBatchCallback.selector,
+            abi.encode(info, callbackFunction, extraData)
+        );
```
TateB, line 306: calling to a resolver compiled with solidity <=0.4.11 with an unsupported method here will use all the gas of the call once it hits in the batch, in prior URs we did an interface check for each, and if unsupported capped the gas for a call at 50k. try test with `jessesum.eth`

test/universalResolver/TestUniversalResolver.ts
```diff
@@ -0,0 +1,627 @@
...
+        dnsEncodeName(testName),
+        bundle.call,
+      ])
+      expectVar({ resolver }).toEqualAddress(F.Shapeshift1.address)
+      expectVar({ answer }).toStrictEqual(bundle.answer)
+      bundle.expect(answer)
+    })
+  })
+
+  describe('reverse()', () => {
```
TateB, line 489: worth testing non COIN_TYPE_ETH since it's a different codepath

test/universalResolver/TestUniversalResolver.ts
```diff
@@ -0,0 +1,627 @@
...
+      await F.Shapeshift1.write.setOffchain([true])
+      const [answer, resolver] = await F.UniversalResolver.read.resolve([
+        dnsEncodeName(testName),
+        bundle.call,
+      ])
+      expectVar({ resolver }).toEqualAddress(F.Shapeshift1.address)
+      expectVar({ answer }).toStrictEqual(bundle.answer)
+      bundle.expect(answer)
+    })
+  })
```
TateB, line 487: test for onchain w/multicall - 1 reverting call + 1 success

TateB (changes requested):
looking very clean, just minor tweaks/issues

TateB (approved):

PR #428, opened 2025-04-06
fix: NameCoder disallow literal-0 hashed label

adraffy:
- change to revert on `[0000000000000000000000000000000000000000000000000000000000000000]`
- added test


Arachnid (commented):
What's the motivation for this change?

adraffy:
I was thinking that if `readLabel()` returns `0` it should imply the root (terminal) node.

Arachnid (approved):

Arachnid:
Good point. Can you add a comment to that effect? Something like "Disallow non-zero-length labels with a labelhash of 0 to prevent confusion with DNS terminator".

PR #429, opened 2025-04-10
Add `contenthash()` and `.write` to universalResolver/utils.ts

adraffy:
- added `KnownProfile.contenthash` to get/set`contenthash()`
- added `Expected.write` for use in setters
      - `KnownBundle.write` is `multicall()`

---

The setters weren't necessary for `UR` testing, as `DummyShapeshifterResolver` doesn't use setters, but `PublicResolver` and  namechain `OwnedResolver` both use setters.

### DIFF
--- a/test/universalResolver/TestUniversalResolver.ts
+++ b/test/universalResolver/TestUniversalResolver.ts
@@ -44,7 +44,7 @@ const resolutions = makeResolutions({
   addresses: [
     {
       coinType: COIN_TYPE_ETH,
-      encodedAddress: anotherAddress,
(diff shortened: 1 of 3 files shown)


Mrt3383x38:
Artık şu sorunu cözüme bağlayıp işimize bakalım 4 yıldır maduruz bide kendi parAmızın peşinde kosuyoruz

adraffy:
Probably better to just replace this [with the version](https://github.com/ensdomains/namechain/blob/main/contracts/test/utils/resolutions.ts) in namechain repo which supports all resolver profiles.

PR #432, opened 2025-04-30
add `DefaultReverseResolver`, `ETHReverseResolver`, redesign `ChainReverseResolver`

adraffy:
* added `{contracts,test}/reverseResolver/` directories
* added [AbstractReverseResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/contracts/reverseResolver/AbstractReverseResolver.sol)
    * has `coinType()`, `chainId()`
    * answers `addr(coinType()) = registrar` for namespace
    * added subclass [DefaultReverseResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/contracts/reverseResolver/DefaultReverseResolver.sol)
        * should be set on `reverse` (and maybe `default.reverse` but unnecessary)
             * also works on any `{evmCoinType}.reverse` and intercept all EVM coin types
        * added [TestDefaultReverseResolver.ts](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/test/reverseResolver/TestDefaultReverseResolver.ts)
    * added subclass [ETHReverseResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/contracts/reverseResolver/ETHReverseResolver.sol) (for V2)
        * enables V2 registrar with V1 fallback
        * added [TestETHReverseResolver.ts](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/test/reverseResolver/TestETHReverseResolver.ts)
    * added subclass [ChainReverseResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/contracts/reverseResolver/ChainReverseResolver.sol) 
       * added [TestChainReverseResolver.ts](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/test/reverseResolver/TestChainReverseResolver.ts)
* changed `ownedEnsFixture.ts` into just [deployRegistryFixture.ts](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/test/fixtures/deployRegistryFixture.ts), move rest of UR-specific fixture into tests
* added [deployDefaultReverseFixture.ts](https://github.com/ensdomains/ens-contracts/blob/feature/default-reverse-resolver/test/fixtures/deployDefaultReverseFixture.ts)
[...]

PR #436, opened 2025-05-08
Relax `HexUtils` to support unpadded, add `ENSIP19.parse()`

adraffy:
* changed [HexUtils.sol](https://github.com/ensdomains/ens-contracts/blob/feat/bet-368-relax-hex/contracts/utils/HexUtils.sol) to support unpadded hex and not revert unnecessarily 
    * uses right alignment so `0x100` is `0x0100`
    * all callsites respect `valid` so results in better errors
    * more gas efficient
    * updated [TestHexUtils.ts](https://github.com/ensdomains/ens-contracts/blob/feat/bet-368-relax-hex/test/utils/TestHexUtils.ts) with expanded test coverage
* added `parse()` and `isEVMCoinType()` to [ENSIP19.sol](https://github.com/ensdomains/ens-contracts/blob/feat/bet-368-relax-hex/contracts/utils/ENSIP19.sol)
    * updated [TestENSIP19.ts](https://github.com/ensdomains/ens-contracts/blob/feat/bet-368-relax-hex/test/utils/TestENSIP19.ts)

### DIFF
--- a/contracts/utils/ENSIP19.sol
+++ b/contracts/utils/ENSIP19.sol
@@ -67,18 +67,6 @@ library ENSIP19 {
         );
     }
 
-    /// @dev Same as `reverseName()` but uses EVM Address + Chain ID.
-    function reverseName(
-        address addr,
-        uint64 chain
-    ) internal pure returns (string memory) {
-        return
-            reverseName(
-                abi.encodePacked(addr),
-                chain == CHAIN_ID_ETH ? COIN_TYPE_ETH : chain | EVM_BIT
-            );
-    }
-
     /// @dev Parse Reverse Name into Encoded Address + Coin Type.
     ///      Matches: /^[0-9a-fA-F]+\.([0-9a-f]{1,64}|addr|default)\.reverse$/.
     ///      Reverts `DNSDecodingFailed`.
--- a/contracts/utils/ENSIP19.sol
+++ b/contracts/utils/ENSIP19.sol
@@ -42,18 +42,6 @@ library ENSIP19 {
         isEVM = chainFromCoinType(coinType) != 0 || coinType == EVM_BIT;
     }
 
-    /// @dev Same as `reverseName()` but uses EVM Address + Chain ID.
-    function reverseName(
-        address addr,
-        uint64 chain
-    ) internal pure returns (string memory) {
-        return
-            reverseName(
-                abi.encodePacked(addr),
-                chain == CHAIN_ID_ETH ? COIN_TYPE_ETH : chain | EVM_BIT
-            );
-    }
-
     /// @dev Generate Reverse Name from Encoded Address + Coin Type.
     ///      Reverts `EmptyAddress` if `encodedAddress` is `0x`.
     /// @param encodedAddress The input address.
@@ -77,6 +65,18 @@ library ENSIP19 {
                 TLD_REVERSE
             )
         );
+    }
+
+    /// @dev Same as `reverseName()` but uses EVM Address + Chain ID.
+    function reverseName(
+        address addr,
+        uint64 chain
+    ) internal pure returns (string memory) {
+        return
+            reverseName(
+                abi.encodePacked(addr),
+                chain == CHAIN_ID_ETH ? COIN_TYPE_ETH : chain | EVM_BIT
+            );
     }
 
     /// @dev Parse Reverse Name into Encoded Address + Coin Type.
--- a/contracts/utils/ENSIP19.sol
+++ b/contracts/utils/ENSIP19.sol
@@ -42,19 +42,19 @@ library ENSIP19 {
         isEVM = chainFromCoinType(coinType) != 0 || coinType == EVM_BIT;
     }
 
-    /// @dev Generate Reverse Name from Encoded Address + Coin Type.
-    ///      Reverts `EmptyAddress` if `encodedAddress` is `0x`.
-    /// @param encodedAddress The input address.
+    /// @dev Generate Reverse Name from Address + Coin Type.
+    ///      Reverts `EmptyAddress` if `addressBytes` is `0x`.
+    /// @param addressBytes The input address.
     /// @param coinType The coin type.
     /// @return name The ENS reverse name, eg. `1234abcd.addr.reverse`.
     function reverseName(
-        bytes memory encodedAddress,
+        bytes memory addressBytes,
         uint256 coinType
     ) internal pure returns (string memory name) {
-        if (encodedAddress.length == 0) revert EmptyAddress();
+        if (addressBytes.length == 0) revert EmptyAddress();
         name = string(
             abi.encodePacked(
-                HexUtils.bytesToHex(encodedAddress),
+                HexUtils.bytesToHex(addressBytes),
                 bytes1("."),
                 coinType == COIN_TYPE_ETH
                     ? SLUG_ETH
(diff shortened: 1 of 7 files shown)


contracts/utils/ENSIP19.sol
```diff
@@ -27,16 +33,13 @@ library ENSIP19 {
...
-                abi.encodePacked(addr),
-                chain == 1 ? COIN_TYPE_ETH : chain | EVM_BIT
-            );
+    /// @dev Determine if Coin Type is for an EVM address.
+    /// @param coinType The coin type.
+    /// @return isEVM True if coin type represents an EVM address.
+    function isEVMCoinType(
+        uint256 coinType
+    ) internal pure returns (bool isEVM) {
+        isEVM = chainFromCoinType(coinType) != 0 || coinType == EVM_BIT;
```
Arachnid, line 42:
```suggestion
        isEVM = coinType & EVM_BIT == EVM_BIT;
```

contracts/utils/HexUtils.sol
```diff
@@ -2,45 +2,90 @@
...
+    function hexToAddress(
+        bytes memory hexString,
+        uint256 pos,
+        uint256 end
+    ) internal pure returns (address addr, bool valid) {
+        if (end - pos < 40) return (address(0), false); // too short
+        bytes32 word;
+        (word, valid) = hexStringToBytes32(hexString, pos, end);
+        addr = address(uint160(uint256(word)));
+    }
```
Arachnid: Does it make any sense to have this function? If it's >40 bytes, the high bytes will be silently discarded.

contracts/utils/ENSIP19.sol
```diff
@@ -52,14 +55,64 @@ library ENSIP19 {
...
+            );
+    }
+
+    /// @dev Parse Reverse Name into Encoded Address + Coin Type.
+    ///      Matches: /^[0-9a-fA-F]+\.([0-9a-f]{1,64}|addr|default)\.reverse$/.
+    ///      Reverts `DNSDecodingFailed`.
+    /// @param name The DNS-encoded name.
+    /// @return encodedAddress The address or empty if invalid.
+    /// @return coinType The coin type.
+    function parse(
```
Arachnid, line 76: What is this for? It doesn't seem to be used anywhere.

Arachnid (changes requested):

contracts/utils/ENSIP19.sol
adraffy: I want to use it in `L1ReverseResolver` and `L1DefaultReverseResolver`.  I was worried that PR was too complex since it did `ReverseRegistrar + ReverseResolver + <this pr>`.

tate recently refactored out the registrar changes and I refactored out these changes from [this PR](https://github.com/ensdomains/ens-contracts/pull/432).

contracts/utils/ENSIP19.sol
adraffy: Oh yeah, good point

adraffy: wait no, this doesn't cover `coinType = 60`, and gives false positives on large/exotic coin types.

contracts/utils/HexUtils.sol
adraffy: I kept the function behavior from before.  I figured it was just the companion to `addressToHex()`.

The 40 byte guard could be removed but this function does define the alignment, eg. `uint160` (right) not `bytes20` (left).

contracts/utils/ENSIP19.sol
adraffy: Table from [prior comment](https://github.com/ensdomains/ensips/pull/7#issuecomment-2845908779)

| Name | CoinType | `chainFromCoinType()` | `isEVMCoinType()` | `reverseName()` | 
| -- | --: | --: | -- | -- | 
| Bitcoin | `0` | `0` | `false` | `"0.reverse"` | 
| Ethereum | `60` | `1` | `true` | `"addr.reverse"` | 
| *default* | `0x8000_0000` | `0` | `true` | `"default.reverse"` |
| Base | `0x8000_2105`  |  `8453` | `true` | `"80002105.reverse"` | 
| *random exotic* | `0x1_8000_0123`  |  `0` | `false` | `"180000123.reverse"` |

contracts/utils/ENSIP19.sol
adraffy: It's also the inverse of `reverseName()`:
`ENSIP19.parse(NameCoder.encode(ENSIP19.reverseName(a, c))) == (a, c)`

The reason for unpadded parsing was so I could parse the coin type from a reverse name, which follows `Number.toString(16)` conventions.

contracts/utils/ENSIP19.sol
adraffy: `coinType & ~(EVM_BIT-1) == EVM_BIT || coinType == 60` would work but I think the original was more clear.

contracts/utils/HexUtils.sol
Arachnid: I think this function is unsafe as-is - it will only work safely for 40 byte addresses, so that's all we should support.

contracts/utils/ENSIP19.sol
Arachnid: I think `coinType & EVM_BIT == EVM_BIT || coinType = 60` would work fine here, since we don't support coin types over 31 bits at present.

Arachnid (changes requested):
As per comments

contracts/utils/HexUtils.sol
adraffy: Just to be clear, the only change is odd-length support:
```                
                <-------------- 40 chars -------------->
hexToAddress(  "0000000000000000000000000000000000000001") = address(1) 
hexToAddress( "F0000000000000000000000000000000000000001") = address(1) // odd, failed before
hexToAddress("FF0000000000000000000000000000000000000001") = address(1)
```

So are you saying change to exactly 40 bytes? (instead of 40-64)

Arachnid: Good catch. Yes, I think we have to do that - silently dropping address bits is unsafe.

Arachnid (changes requested):

contracts/utils/ENSIP19.sol
adraffy: Recapping Slack discussion:

* coin types with embedded chain ids above `0x7FFFFFFF` are undefined
* `isEVMCoinType()` and `chainFromCoinType()` only respect embedded chain ids that fit in 32 bits (upper 224 bits must be zero)
* `chain id` is defined as `uint64` in the EVM
* [EIP-7785](https://github.com/ethereum/ERCs/blob/master/ERCS/erc-7785.md) introduces 256-bit chain ids
* [Possibly](https://ethereum-magicians.org/t/erc-7785-onchain-registration-of-chain-identifiers/21299/7?u=adraffy) 7785 could have a recommendation that the 31th bit is forced (effectively making it a 255-bit chain id) so that we can differentiate wide evm coin types from others

I also removed the overloaded `reverseName()` that used chain id to avoid any confusion.

contracts/utils/ENSIP19.sol
```diff
@@ -52,14 +55,52 @@ library ENSIP19 {
...
+        bytes memory name
+    ) internal pure returns (bytes memory encodedAddress, uint256 coinType) {
```
TateB: this is a little confusing since encodedAddress in a multicoin context refers to the encoded human readable representation of an address, which this is not (it's kinda the opposite). not too relevant but worth noting

TateB (approved):

contracts/utils/ENSIP19.sol
adraffy: What do you call it?  Raw address?

I think I always found the `address-encoder` terminology backwards.  I consider `encode()` the byte-ification.

adraffy: Changed to `addressBytes` after discussion.

Arachnid (approved):

PR #438, opened 2025-05-20
feat: new controller

TateB:


contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -38,102 +28,113 @@ contract ETHRegistrarController is
...
     uint64 private constant MAX_EXPIRY = type(uint64).max;
+
+    ENS public immutable ens;
+
     BaseRegistrarImplementation immutable base;
+
     IPriceOracle public immutable prices;
+
     uint256 public immutable minCommitmentAge;
+
```
Arachnid: What's with all the extra whitespace?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -144,84 +145,128 @@ contract ETHRegistrarController is
...
+                    msg.sender,
+                    msg.sender,
+                    registration.resolver,
+                    string.concat(registration.label, ".eth")
+                );
+            else if (registration.reverseRecord == ReverseRecord.Default)
+                defaultReverseRegistrar.setNameForAddr(
+                    msg.sender,
+                    string.concat(registration.label, ".eth")
+                );
```
Arachnid, line 425: Won't this provide a route where a contract could end up setting a default reverse record?

Arachnid (changes requested):
While we're making a new controller anyway, could you add a function that allows the owner of the controller (the DAO) to register a name without payment? It should bypass the name length and payment checks but still enforce all other restrictions.

Arachnid:
@TateB Let's make the price oracle mutable too, so any pricing changes won't require another new controller.

contracts/ethregistrar/ETHRegistrarController.sol
TateB: that was meant to be for comments but i forgot to add them... added

contracts/ethregistrar/ETHRegistrarController.sol
TateB: yes, the limitation of EOAs only for the default namespace is a client-side distinction (mainly because of 7702)

Arachnid: I'm not sure I follow. Isn't it a risk if a contract sets a default record for itself, since that address could be used by another contract elsewhere? Shouldn't we do what we can to prevent this?

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,31 +1,21 @@
 //SPDX-License-Identifier: MIT
 pragma solidity ~0.8.17;
 
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
```
adraffy: I think `Address` can be removed.  `IERC165` can be proper `ERC165` (this applies to other registrar contracts too.)

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -37,197 +27,296 @@ contract ETHRegistrarController is
...
         BaseRegistrarImplementation _base,
         IPriceOracle _prices,
         uint256 _minCommitmentAge,
         uint256 _maxCommitmentAge,
-        ReverseRegistrar _reverseRegistrar,
-        INameWrapper _nameWrapper,
+        IReverseRegistrar _reverseRegistrar,
+        IDefaultReverseRegistrar _defaultReverseRegistrar,
         ENS _ens
     ) ReverseClaimer(_ens, msg.sender) {
```
adraffy: Since we have `Ownable` reverse support and the controller is ownable, should we remove `ReverseClaimer`?

contracts/ethregistrar/ETHRegistrarController.sol
adraffy: What would be the solution?  Require a signature for this case?  Change `.reverseRecord` to `bytes` and use `bytes(0)` as `.None` and `bytes(1)` as `.Ethereum`.

Or is there some 7702 detection method?

adraffy (changes requested):

contracts/ethregistrar/ETHRegistrarController.sol
Arachnid: Requiring a signature seems the simplest option, and doing it that way, the contract wouldn't even have to be aware of the new controller.

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -37,197 +27,296 @@ contract ETHRegistrarController is
...
+    ///
+    /// @param registration The registration to register.
+    /// @param registration.label The label of the name.
+    /// @param registration.owner The owner of the name.
+    /// @param registration.duration The duration of the registration.
+    /// @param registration.resolver The resolver for the name.
+    /// @param registration.data The data for the name.
+    /// @param registration.reverseRecord The reverse record for the name.
+    /// @param registration.referrer The referrer of the registration.
+    function registerWithoutPayment(
```
Arachnid, line 302: This function also shouldn't require a valid commitment, since it's going to be onchain for a week before it's executed!

Arachnid (changes requested):

contracts/ethregistrar/ETHRegistrarController.sol
TateB: i'm not sure that risk is relevant given that a user/dev would need to go out of their way to set a default record for a contract in the first place? if they (as a developer interacting with a contract directly or through their own code) want to take that risk, it's on them. the risk that a user deploying a contract on another chain unintentionally runs into this is low, but even then, they can just set the chain-specific reverse and resolve the issue.

contracts/ethregistrar/ETHRegistrarController.sol
TateB: i'll use erc165 but the ierc165 import is still needed because the natspec for `supportsInterface` is just `See IERC165.supportsInterface` which is useless

TateB:
- removed reverseclaimer
- removed import for oz `Address`
- removed commitment requirement for `registerWithoutPayment`

Arachnid:
Please push your changes so I can review them!

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,289 +1,431 @@
...
+        Registration calldata registration
+    ) public onlyOwner {
+        bytes32 labelhash = keccak256(bytes(registration.label));
+
+        if (!base.available(uint256(labelhash)))
+            revert NameNotAvailable(registration.label);
```
Arachnid, line 308: Can this check be moved into `_register`?

Arachnid (approved):

contracts/ethregistrar/ETHRegistrarController.sol
```diff
@@ -1,289 +1,438 @@
...
+    ///
+    /// @param registration The registration to register.
+    /// @param registration.label The label of the name.
+    /// @param registration.owner The owner of the name.
+    /// @param registration.duration The duration of the registration.
+    /// @param registration.resolver The resolver for the name.
+    /// @param registration.data The data for the name.
+    /// @param registration.reverseRecord The reverse record for the name.
+    /// @param registration.referrer The referrer of the registration.
+    function registerWithoutPayment(
```
clowestab, line 302: Having thought about this further, I feel as though `registerWithoutPayment` should be implemented within a separate controller authorised on the `BaseRegistrarImplementation`.

This would maintain a clear separation of concerns whilst making registrations through this alternative path more easily auditable.

It also means that if the DAO changes its positioning down the line this functionality can be removed more easily.

meny3437-blip (commented):
Controller @meny3437-blip

PR #442, opened 2025-06-03
Add default fallback to `AddrResolver`

adraffy:
- added default fallback to [AddrResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-383-addr-resolver/contracts/resolvers/profiles/AddrResolver.sol)
     - uses same logic as [DedicatedResolver.sol](https://github.com/ensdomains/namechain/blob/feat/bet-371-dedicated-resolver/contracts/src/common/DedicatedResolver.sol)
- added [IHasAddressResolver.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-383-addr-resolver/contracts/resolvers/profiles/IHasAddressResolver.sol)
- added corresponding [tests](https://github.com/ensdomains/ens-contracts/blob/feature/bet-383-addr-resolver/test/resolvers/TestPublicResolver.ts)
- changed `EVM_BIT` to `COIN_TYPE_DEFAULT`

### DIFF
--- a/contracts/resolvers/profiles/AddrResolver.sol
+++ b/contracts/resolvers/profiles/AddrResolver.sol
@@ -1,62 +1,24 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.4;
 
-import "../ResolverBase.sol";
-import "./IAddrResolver.sol";
-import "./IAddressResolver.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {ResolverBase} from "../ResolverBase.sol";
+import {IAddrResolver} from "./IAddrResolver.sol";
+import {IAddressResolver} from "./IAddressResolver.sol";
+import {ENSIP19, COIN_TYPE_ETH, EVM_BIT} from "../../utils/ENSIP19.sol";
 
 abstract contract AddrResolver is
     IAddrResolver,
     IAddressResolver,
     ResolverBase
 {
-    uint256 private constant COIN_TYPE_ETH = 60;
-
     mapping(uint64 => mapping(bytes32 => mapping(uint256 => bytes))) versionable_addresses;
 
-    /// Sets the address associated with an ENS node.
-    /// May only be called by the owner of that node in the ENS registry.
-    /// @param node The node to update.
-    /// @param a The address to set.
-    function setAddr(
-        bytes32 node,
-        address a
-    ) external virtual authorised(node) {
-        setAddr(node, COIN_TYPE_ETH, addressToBytes(a));
-    }
+    /// @notice The supplied address could not be converted to `address`.
+    /// @dev Error selector: `0x8d666f60`
+    error InvalidEVMAddress(bytes addressBytes);
 
-    /// Returns the address associated with an ENS node.
-    /// @param node The ENS node to query.
-    /// @return The associated address.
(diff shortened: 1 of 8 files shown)


contracts/resolvers/profiles/AddrResolver.sol
```diff
@@ -1,84 +1,113 @@
...
     mapping(uint64 => mapping(bytes32 => mapping(uint256 => bytes))) versionable_addresses;
 
-    /// Sets the address associated with an ENS node.
-    /// May only be called by the owner of that node in the ENS registry.
+    /// @notice The supplied address could not be converted to `address`.
+    /// @dev Error selector: `0x8d666f60`
+    error InvalidEVMAddress(bytes addressBytes);
+
+    /// @inheritdoc IERC165
+    function supportsInterface(
```
TateB: nit: supportsInterface at bottom

contracts/resolvers/profiles/AddrResolver.sol
```diff
@@ -1,84 +1,113 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.4;
 
-import "../ResolverBase.sol";
-import "./IAddrResolver.sol";
-import "./IAddressResolver.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
```
TateB: nit: empty line between external and internal deps

TateB (changes requested):

contracts/resolvers/profiles/AddrResolver.sol
```diff
@@ -1,84 +1,112 @@
...
     function setAddr(
         bytes32 node,
-        address a
+        address _addr
     ) external virtual authorised(node) {
-        setAddr(node, COIN_TYPE_ETH, addressToBytes(a));
+        setAddr(
+            node,
+            COIN_TYPE_ETH,
+            _addr == address(0) ? bytes("") : abi.encodePacked(_addr)
```
Arachnid: I think it's probably less surprising to have this retain its existing behaviour, and have users call the overload that takes `bytes` if they want to explicitly delete the address.

contracts/resolvers/profiles/AddrResolver.sol
```diff
@@ -1,84 +1,112 @@
...
-        return versionable_addresses[recordVersions[node]][node][coinType];
+    ) public view virtual override returns (bytes memory addressBytes) {
+        mapping(uint256 => bytes) storage addrs = versionable_addresses[
+            recordVersions[node]
+        ][node];
+        addressBytes = addrs[coinType];
+        if (
+            addressBytes.length == 0 && ENSIP19.chainFromCoinType(coinType) > 0
+        ) {
+            addressBytes = addrs[EVM_BIT];
```
Arachnid: I think it'd be clearer if we defined a new constant `CHAIN_ID_ZERO` for this purpose.

Arachnid (changes requested):

contracts/resolvers/profiles/AddrResolver.sol
adraffy: Discussed in Slack, using `COIN_TYPE_DEFAULT`

Arachnid (approved):

PR #448, opened 2025-06-12
fix: deploy migrations

TateB:
- added IDs for all deploy scripts
  - this allows all deploy scripts to be uniquely identified, and won't run again after being run once (based on migrations list)
- added `true` returns for deploy scripts, meaning their ID is logged in migrations
- added unique deploy IDs for each `ChainReverseResolver` deployment
- added a tenderly proxy script, which allows pseudo account impersonations (tenderly virtual testnet authorises all addresses) through capturing `eth_accounts` calls


jefflau (changes requested):
Can you add something into the readme to explain the versioning so anyone else that deploys knows what they need to increment to make sure things work properly

Arachnid (approved):

scripts/deploy-with-impersonation.ts
Arachnid, line 1: Can you add some kind of documentation on how/when/why to use this?

PR #449, opened 2025-06-12
Fix metadata service URL in Name Wrapper

gskril:


TateB (approved):

gskril:
I was wrong; the currently deployed `uri` of `https://metadata.ens.domains/mainnet/0xD4416b13d2b3a9aBae7AcD5D6C2BbDBE25686401/0x{id}` is formatted correctly.

I assumed `{id}` was the number but according to the spec:

> The uri can include the string {id} which clients must replace with the actual token ID, in lowercase hexadecimal (with no 0x prefix) and leading zero padded to 64 hex characters.

As an example: this would result in `https://metadata.ens.domains/mainnet/0xD4416b13d2b3a9aBae7AcD5D6C2BbDBE25686401/0x483307b4efe2e6536131000b3f20cba919c05e4d39b4fde6d43384167b31560d` for `gtest.eth`, which is correct.

PR #458, opened 2025-07-03
Added controllers in L2 reverse registrar

NickSneo:
Before ENSIP-19 is merged to mainnet, we’d like to propose a small edit.

We suggest adding support for controllers in the `L2ReverseRegistrar` contract. This change would allow us to inject additional logic for determining which smart contracts are eligible to set a primary ENS name.

Currently, only contracts that implement `Ownable` can set a primary name, due to the ownership check which is hardcoded for contracts. By introducing controllers, we can expand support and enable more contracts to qualify for a primary name.

These controllers could support various use cases (still open for discussion), such as:
	•	Allowing only a specific set of contracts to be named
	•	Allowing verified deployers to set primary names
	•	Supporting contracts deployed via Account Abstraction (AA), etc.

Our goal is to ensure that before ENSIP-19 is finalized, we retain the flexibility to define additional rules—so more contracts, both existing and future, can be supported with primary ENS names.

### DIFF
--- a/contracts/reverseRegistrar/L2ReverseRegistrar.sol
+++ b/contracts/reverseRegistrar/L2ReverseRegistrar.sol
@@ -15,13 +15,23 @@ import {SignatureUtils} from "./SignatureUtils.sol";
 contract L2ReverseRegistrar is
     IL2ReverseRegistrar,
     ERC165,
-    StandaloneReverseRegistrar
+    StandaloneReverseRegistrar,
+    Ownable
 {
     using SignatureUtils for bytes;
     using MessageHashUtils for bytes32;
 
     /// @notice The coin type for the chain this contract is deployed to.
     uint256 public immutable coinType;
+
+    /// @notice A map of addresses that are authorised to set names for any address.
+    mapping(address => bool) public controllers;
+
+    /// @notice Emitted when a controller is added.
+    event ControllerAdded(address indexed controller);
+
+    /// @notice Emitted when a controller is removed.
+    event ControllerRemoved(address indexed controller);
 
     /// @notice Thrown when the specified address is not the owner of the contract
     error NotOwnerOfContract();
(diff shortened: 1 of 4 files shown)


adraffy (commented):
I agree with the [idea](https://discuss.ens.domains/t/approaches-to-support-setting-primary-names-for-all-contracts/20919/4?u=raffy) but we need a solution for DAO-approved execution on other chains as there currently is no "owner" and choice of registrar is guaranteed by the resolver on L1, which is controlled by the DAO.

One idea would be: the DAO could execute crosschain messages for each rollup that targets a relayer contract on each chain.  The registrar on each chain would need to trust the corresponding relayer.  The registrar would need to be deployed to each chain with the corresponding relayer as the initial controller.  The relayer would forward or claim messages from the messaging service, verify they came from the DAO wallet, and execute the calldata.  Breaking changes to the messaging service could be mitigated by adding a new relayer as an additional controller.

```solidity
// L2ReverseRegistrar.sol
constructor(..., address firstController) {
   controllers[firstController] = true;
   ...
}
function addController(address controller) external onlyController { ... }
function removeController(address controller) external onlyController { ... }

// L2ENSDAORelayer.sol
function claimMessage(...) {
    // ensure from = DAO
    // claim from messaging service
    // relay execution
}
```

Arachnid:
Thanks for the contribution!

We've discussed this at length internally. It's a good idea, and would provide more flexibility to how the reverse registrar operates. However, it also requires having an owner account on each chain,whereas the current code has no privileged roles. Doing this cross-chain is not straightforward: we'd need to establish a relayer contract on each chain, which can receive messages from either the official bridge for that chain or from a third-party service such as CCIP, and configure them such that the DAO can execute proposals on L1 to generate cross-chain messages on each L2.

Doing this now adds a lot of complexity and would definitely delay the rollout of primary names on L2s. Instead, we think it's wiser to stick with the current implementation, and if/when this needs upgrading, we can roll out a new registrar that falls back to looking data up in the old one. The new registrar can incorporate this functionality if needed.

PR #459, opened 2025-07-19
`NameCoder` Helpers

adraffy:
- added [nextLabel()](https://github.com/ensdomains/ens-contracts/blob/fix/namecoder-iter/contracts/utils/NameCoder.sol) and `prevLabel()`
- added `namehash(bytes32, bytes32) returns (bytes32)`
- added `readLabel()` w/optional hashed label support
- added `matchSuffix()` from `namechain/`
- moved "junk at end" revert from `namehash()` to `nextLabel()`
- added [more tests](https://github.com/ensdomains/ens-contracts/blob/fix/namecoder-iter/test/utils/TestNameCoder.ts)
- changed `RegistryUtils.findResolver()` to use `NameCoder.namehash()`

### DIFF
--- a/contracts/universalResolver/RegistryUtils.sol
+++ b/contracts/universalResolver/RegistryUtils.sol
@@ -29,7 +29,7 @@ library RegistryUtils {
                 bytes32 parentNode,
                 uint256 parentOffset
             ) = findResolver(registry, name, next);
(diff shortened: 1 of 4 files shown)


TateB (approved):

meny3437-blip (commented):
New @meny3437-blip

PR #416, opened 2025-03-09
new UniversalResolver + forge Rewrites

adraffy:
- [IUniversalResolver.sol](https://github.com/ensdomains/ens-contracts/pull/417) PR 
- [UniversalResolver2.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/contracts/universalResolver/UniversalResolver2.sol) inspired from [Unruggable UR](https://github.com/unruggable-labs/unruggable-resolve) design
     - ownership only for `setForwardResolution()` (eg. v1 &rarr; v2)
     - reverse resolution logic is the same for v1 and v2
- [ForwardResolution.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/contracts/universalResolver/ForwardResolution.sol) implements standard forward resolution for v1
    - another version implementing [IForwardResolution.sol](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/contracts/universalResolver/IForwardResolution.sol) can be written for v2
    - ownership only for `setBatchedGateways()`
    - if `resolve(multicall)` is understood by the resolver, the batched gateway is bypassed
         - essentially functions as a version check
         - will use this for 1+ records (instead of 2+)
    - also using a [temporary hack](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/contracts/universalResolver/ForwardResolution.sol#L98-L114) to workaround Coinbase's resolvers
- mainnet tests (`bun run test:remote`)
    - [list of tests](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/test/universalResolver/mainnetTests.ts)
    - [TestForwardResolution.remote.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/test/universalResolver/TestForwardResolution.remote.ts)
    - [TestUniversalResolver2.remote.ts](https://github.com/ensdomains/ens-contracts/blob/feature/bet-193-develop-a-standardized-universalresolver-compatible-with-v1/test/universalResolver/TestUniversalResolver2.remote.ts)
- other tests
[...]

PR #461, opened 2025-07-21
Feature/fet 1977 alternative migrate tests to foundry

mdtanrikulu:


adraffy:
Crazy!

I do think we should use foundry when possible but complex tests should stay in TypeScript as the equivalent Solidity is unreadable.  I think `namechain/` is pretty good: basically anything that isn't CCIP-Read is a foundry test.

Also, I made an earlier [attempt at this](https://github.com/ensdomains/ens-contracts/pull/416/files#diff-5a6f2aa8fa60fbaade8a41eaac5f19bbf0809434312edeed58d47d95514991cc).  Not sure if any of my utility tests are useful.

mdtanrikulu:
> Crazy!
> 
> I do think we should use foundry when possible but complex tests should stay in TypeScript as the equivalent Solidity is unreadable. I think `namechain/` is pretty good: basically anything that isn't CCIP-Read is a foundry test.
> 
> Also, I made an earlier [attempt at this](https://github.com/ensdomains/ens-contracts/pull/416/files#diff-5a6f2aa8fa60fbaade8a41eaac5f19bbf0809434312edeed58d47d95514991cc). Not sure if any of my utility tests are useful.

Sure, I think it will be better in the long run to keep both in the same structure. I believe it's still possible to achieve clear ccip-read foundry tests using ffi functionality but that's just preference. I am happy with both.

TateB:
is it actually useful to have all the pre-existing tests in foundry?

mdtanrikulu:
> is it actually useful to have all the pre-existing tests in foundry?

In my opinion, maintaining test dependencies with Hardhat and its plugins introduces additional complexity, especially when dealing with compatibility issues or required patches. Since many of these concerns are already handled well in Foundry, this could be a good opportunity to consider a gradual transition. We can retain the existing *.ts tests as legacyTests to avoid confusion from a sudden removal.

v1rtl:
> is it actually useful to have all the pre-existing tests in foundry?

yes it is, for test environment. it is planned for tenv to be able to deploy smart contracts without requiring a complex setup involving multiple hardhat versions and other crap

after ens-contracts get migrated to foundry, migrating deploy scripts from ensjs to tenv will be easier and require less setup

.github/workflows/main.yml
```diff
@@ -25,13 +27,15 @@ jobs:
...
+        with:
+          submodules: recursive
 
       - name: Use Bun 1.2.0
         uses: oven-sh/setup-bun@v2
         with:
           bun-version: 1.2.0
 
-      - run: bun install --frozen-lockfile
+      - run: bun install
```
v1rtl, line 38: frozen lockfiles ensure no random transititves get pulled and that the build is reproducible. I suggest to put it back and regenerate the lockfile and commit it to ensure it matches the same one in CI

v1rtl (changes requested):

TateB:
> yes it is, for test environment. it is planned for tenv to be able to deploy smart contracts without requiring a complex setup involving multiple hardhat versions and other crap
> 
> after ens-contracts get migrated to foundry, migrating deploy scripts from ensjs to tenv will be easier and require less setup

new deploy scripts are utilising rocketh anyway, how are foundry tests related?

v1rtl:
> > yes it is, for test environment. it is planned for tenv to be able to deploy smart contracts without requiring a complex setup involving multiple hardhat versions and other crap
> > after ens-contracts get migrated to foundry, migrating deploy scripts from ensjs to tenv will be easier and require less setup
> 
> new deploy scripts are utilising rocketh anyway, how are foundry tests related?

ah right yeah tests are not very related, but deploy scripts are

PR #462, opened 2025-07-22
Add `RegistryUtils`

adraffy:
* move `findResolver()` into a [library](https://github.com/ensdomains/ens-contracts/blob/fix/registry-utils/contracts/universalResolver/RegistryUtils.sol)
* refactor [UniversalResolver](https://github.com/ensdomains/ens-contracts/blob/fix/registry-utils/contracts/universalResolver/UniversalResolver.sol) to use it
* improve comments


TateB (approved):

meny3437-blip (commented):
Add @meny3437-blip

PR #464, opened 2025-07-25
Make `BytesUtils.sol` Remappable

adraffy:
- moved `BytesUtils.sol` into `BytesUtils/BytesUtils.sol`
- create new [BytesUtils.sol](https://github.com/ensdomains/ens-contracts/blob/fix/remap-bytesutils/contracts/utils/BytesUtils.sol) that imports `BytesUtils/BytesUtils.sol`


adraffy:
https://github.com/foundry-rs/foundry/pull/11184

Not needed with above patch in v1.3.1

PR #477, opened 2025-09-05
Add `IVerifiableGateway`

adraffy:
- added [IVerifiableGateway.sol](https://github.com/ensdomains/ens-contracts/blob/feat/verifiable-gateway/contracts/ccipRead/IVerifiableGateway.sol)
- modified `ChainReverseResolver` to implement `IVerifiableGateway` and `IGatewayProvider`


adraffy:
Using `IVerifiableResolver` instead.

PR #482, opened 2025-09-09
Add `ReverseNamer`

adraffy:
- added [ReverseNamer.sol](https://github.com/ensdomains/ens-contracts/blob/feat/v2-reverse-claimer/contracts/reverseRegistrar/ReverseNamer.sol) and [tests](https://github.com/ensdomains/ens-contracts/blob/feat/v2-reverse-claimer/test/reverseRegistrar/TestReverseNamer.test.ts)
    1. `NamedOnce` for simple immutable name on constructor
    2. `NameableBy` for delegated naming
    3. `ReverseNamer.setName("<name>")` for custom implementation

Usage:
```solidity
// 1
import {NamedOnce} from "@ensdomains/ens-contracts/contracts/reverseRegistrar/ReverseNamer.sol";
contract C is NamedOnce("mycontract.eth") { ... }

// 2
import {NameableBy} from "@ensdomains/ens-contracts/contracts/reverseRegistrar/ReverseNamer.sol";
contract C is NameableBy {
    constructor(address namer) NameableBy(namer, "") { ... }
}

// 3
import {ReverseNamer} from "@ensdomains/ens-contracts/contracts/reverseRegistrar/ReverseNamer.sol";
contract C {
    function f() external {
        ReverseNamer.setName("mycontract.eth");
    }
}
```

---

Alternative ideas:
* `NamedOnce` &rarr; `emit Named(string)` &rarr; generate proof of log
* code snippet: `address(<registrar>).call(abi.encodeWithSignature("setName(string)", "<name>"))`

### DIFF
--- a/contracts/reverseRegistrar/ReverseNamer.sol
+++ b/contracts/reverseRegistrar/ReverseNamer.sol
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IReverseRegistrarStub {
+    function setName(string memory name) external;
+}
+
+/// @notice https://docs.ens.domains/ensip/19/
+library ReverseNamer {
+    address constant REVERSE_REGISTRAR_MAINNET = address(0); // TODO: ENSv2 addr.reverse registrar
+    address constant REVERSE_REGISTRAR_MAINNET_ROLLUP =
+        0x0000000000D8e504002cC26E3Ec46D81971C1664;
+    address constant REVERSE_REGISTRAR_TESTNET_ROLLUP =
+        0x00000BeEF055f7934784D6d81b6BC86665630dbA;
+
+    function registrarFromChain(
+        uint256 chainId
+    ) internal pure returns (address) {
+        if (isMainnet(chainId)) {
+            return REVERSE_REGISTRAR_MAINNET;
+        } else if (isMainnetRollup(chainId)) {
+            return REVERSE_REGISTRAR_MAINNET_ROLLUP;
+        } else if (isTestnetRollup(chainId)) {
+            return REVERSE_REGISTRAR_TESTNET_ROLLUP;
+        } else {
+            return address(0);
+        }
+    }
+
+    function setName(string memory primary) internal {
+        address registrar = registrarFromChain(block.chainid);
+        if (registrar != address(0)) {
+            IReverseRegistrarStub(registrar).setName(primary);
+        }
+    }
+
+    function isMainnet(uint256 chainId) internal pure returns (bool) {
+        return
+            chainId == 1 || // mainnet
+            chainId == 17000 || // holesky
+            chainId == 560048 || // hoodi
+            chainId == 11155111; // sepolia
+    }
+
+    function isMainnetRollup(uint256 chainId) internal pure returns (bool) {
+        return
+            chainId == 10 || // optimism
+            chainId == 8453 || // base
+            chainId == 42161 || // arb1
+            chainId == 59144 || // linea
+            chainId == 534352; // scroll
+    }
+
+    function isTestnetRollup(uint256 chainId) internal pure returns (bool) {
+        return
+            chainId == 59141 || // linea-sepolia
+            chainId == 84532 || // base-sepolia
+            chainId == 421614 || // arb1-sepolia
+            chainId == 534351 || // scroll-sepolia
+            chainId == 11155420; // optimism-sepolia
+    }
+}
+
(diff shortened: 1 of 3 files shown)


gskril:
I like it! Should this PR also delete ReverseClaimer.sol, or maybe more safely just add a natspec comment indicating to use ReverseNamer.sol for wider support (rollups)?

NickSneo:
@adraffy Whats the difference between case 2 and case 3
isn't it indirectly doing same thing delegating someone to name the contract either by directly  delegating in case 2 or indirectly (via some access control on who can call function f()) in case 3

Would it makes sense to just have 2 cases?

contracts/reverseRegistrar/ReverseNamer.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IReverseRegistrarStub {
+    function setName(string memory name) external;
+}
+
+/// @notice https://docs.ens.domains/ensip/19/
+library ReverseNamer {
+    address constant MAINNET = address(0); // TODO: ENSv2 addr.reverse registrar
```
NickSneo, line 10: need to add TESTNET or is it same as MAINNET address? since we have distinct MAINNET_ROLLUP and TESTNET_ROLLUP

contracts/reverseRegistrar/ReverseNamer.sol
```diff
@@ -0,0 +1,99 @@
...
+    function registrarFromChain(
+        uint256 chainId
+    ) internal pure returns (address) {
+        if (isMainnet(chainId)) {
+            return MAINNET;
+        } else if (isMainnetRollup(chainId)) {
+            return MAINNET_ROLLUP;
+        } else if (isTestnetRollup(chainId)) {
+            return TESTNET_ROLLUP;
+        } else {
```
NickSneo, line 25: on line 10 - MAINNET = address(0) so is the else case

contracts/reverseRegistrar/ReverseNamer.sol
```diff
@@ -0,0 +1,99 @@
...
+
+/// @notice Mixin for naming a contract once.
+contract NamedOnce is NotOwnable {
+    constructor(string memory primary) {
+        ReverseNamer.setName(primary);
+    }
+}
+
+/// @notice Mixin for delegated contract naming.
+contract NameableBy is NotOwnable {
```
NickSneo, line 77: NotOwnable in case of NamedOnce makes sense as no one can later name the contract but in case of NameableBy doesnt make much sense to it.

NameableBy can be seen  in a contract with Ownable as an extra delegate who can name contract but still if contract has Owner he can revoke the delegation. I might be wrong but limiting NameableBy to only non ownable contracts doesn't make much sense to me

NickSneo (commented):
PR looks good, thanks for this. We as Enscribe team also need to push more on awareness so developers include this in their contracts before deploying. 

Added a few comments

contracts/reverseRegistrar/ReverseNamer.sol
adraffy: I think this will mean all mainnets (mainnet/sepolia/...).

I speculatively designed [ETHReverseResolver](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/reverseResolver/ETHReverseResolver.sol) as the v2 replacement for `addr.reverse` that uses the same tech as the rollups (independent registry + wildcard resolver) but I think we're going with a proper claimable `{addr}.addr.reverse` that uses the v2 design, but it should be a fixed address deployment and the same on those chains.  And If not, you're correct.

contracts/reverseRegistrar/ReverseNamer.sol
adraffy: Yeah, I wasn't sure about this design.  `NameableBy` could be removed.  My thinking was you want something like `Ownable` but without implementing `Ownable` which at a glance implies the contract is mutable.

`NameableBy + Ownable` would give you 2 accounts that can change the name, but maybe that should be allowed. 

I was thinking of it more like a trust curve: 
1. `NamedOnce` — set once
1. `NameableBy` — set later, revoke it / delegate to someone to name it
1. `NameableBy` — delegate to someone, change whenever
1. `Ownable` — change whenever, but shares with other `Ownable` responsibilities

2pacmcg1 (commented):
JMCG003

adraffy:
Going with https://github.com/ensdomains/contracts-v2/pull/225

PR #486, opened 2025-09-23
Create ETHReferredRenewalProxy.sol

lightwalker-eth:
# Definitions

* Eth2LD: a direct subname of .eth.

# Background

* The "unwrapped" ETHRegistrarController introduced support for tracking a `referrer` for registrations and renewals of Eth2LD.
* There are currently 3 active ETHRegistrarController contracts.
* Only 1 of these active ETHRegistrarController contracts (the "wrapped" ETHRegistrarController) performs a renew operation that keeps expiry state in sync for wrapped Eth2LD between the BaseRegistrar and the NameWrapper.
   * If wrapped Eth2LD are renewed with the other 2 active ETHRegistrarController contracts, the expiry will be updated in the BaseRegistrar but not the NameWrapper.
* Therefore, .eth registrar apps such as the official ENS Manager App as assumed to implement logic to send renewal transactions for wrapped Eth2LD to the "wrapped" ETHRegistrarController and not to use the new "unwrapped" ETHRegistrarController. This use of the "wrapped" ETHRegistrarController causes referral tracking of renewals to be lost.

# Goals
* For all renewals of Eth2LD (wrapped or not), ensure:
   * All referrals emit a `NameRenewed` event that is in full alignment with the `NameRenewed` event emitted by the "unwrapped" ETHRegistrarController.
   * No renewals create an expiry state inconsistency between the BaseRegistrar and the NameWrapper.
* Simplify integration complexity, such that this proxy supports both wrapped and unwrapped Eth2LD. Integrators should not need to check if a Eth2LD is wrapped or not to make use of the proxy.
* Ensure the contract is `Ownable` and deployed such that it can be assigned an ENS primary name.
* Note: It is not a goal to add referral tracking to new registrations of wrapped Eth2LD names.

### DIFF
--- a/contracts/ethregistrar/ETHReferredRenewalProxy.sol
+++ b/contracts/ethregistrar/ETHReferredRenewalProxy.sol
@@ -1,0 +1,169 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ~0.8.17;
+
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {ERC20Recoverable} from "../utils/ERC20Recoverable.sol";
+import {IBaseRegistrar} from "./BaseRegistrarImplementation.sol";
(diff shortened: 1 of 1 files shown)


lightwalker-eth:
Superseded by https://github.com/ensdomains/ens-contracts/pull/487

PR #487, opened 2025-09-23
introduce EthRegistrarControllerWithRenewalReferrals

shrugs:
supersedes https://github.com/ensdomains/ens-contracts/pull/486

```bash
bun run test:remote test/ethregistrar/TestWrappedEthRegistrarControllerWithRenewalReferrals.remote.ts
```

## TODO

- [ ] exhaustive documentation in the contract
- [ ] contract deploy script
- [ ] unwrapped name test is reverting at `ethRegistrarControllerWithReferrals.write.renew` without reason string
- [ ] overpayment test is reverting at `ethRegistrarControllerWithReferrals.write.renew` with obscure "Transaction reverted: function selector was not recognized and there's no fallback nor receive function" 
  - important debugging note: if the price is set exactly (`value: price`) then the renewal passes — it's only _over_ payment that causes this obscure issues
- [ ] the old UniversalResolver remote tests are broken because they haven't been run continuously

## Done

- fixes the repo's management of mainnet forking by creating a dedicated `mainnet-fork` network and `vitest.remote.config.ts` 
- happy path passes — renewals via this wrapper contract are renewed and namewrapper expiry is synced with baseregistar


TateB:
hey, we did just merge a fix for remote testing, but this PR should have local unit testing as well. 
you can deploy the old wrapped registrar in a test by providing the `WrappedETHRegistrarController` mainnet deployment file to `deployContract` as the first arg.

> * happy path passes — renewals via this wrapper contract are renewed and namewrapper expiry is synced with baseregistar

overpayment is actually our happy path. the ETHUSD rate can fluctuate from the time that a transaction is created in a client, and in the time it stays pending in the mempool. to accomodate for this, we add a few % on top to ensure a transaction doesn't fail, and use it as the default in our tests.

> * unwrapped name test is reverting at `ethRegistrarControllerWithReferrals.write.renew` without reason string
> * overpayment test is reverting at `ethRegistrarControllerWithReferrals.write.renew` with obscure

both these errors are happening because there is no `receive`/`fallback` func implemented. it can just be implemented as an empty handler like `receive() external payable { }`

also of note, from some basic gas comparisons the implementation currently seems to be very inefficient. (using a buffered payment)
```
WrappedETHRegistrarController.renew() - 87474 gas used
UnwrappedETHRegistrarController.renew() - 70494 gas used (-19.41%)
EthRegistrarControllerWithRenewalReferrals.renew() - 126640 gas used (+44.77%)
```

shrugs:
yeah the gas costs of this approach effectively runs renew twice, one on each controller. the benefit is that it centralizes the referrer information into the only ens-protocol-native location in the new registrar controller. and also keeps the namewrapper up to date with the new expiry. since this is more of a temporary thing, i'm happy to explore other more gas efficient approaches (namely just using the wrapped controller and emitting a separate event).

shrugs:
just going to close this as outdated

PR #492, opened 2025-10-01
Fix `BaseRegistrarImplementation` warnings

adraffy:


adraffy:
https://getfoundry.sh/config/reference/solidity-compiler/#ignored_warnings_from

PR #495, opened 2025-10-07
fix: multicall3 + other shared deploys

TateB:
separated multicall3 into standalone script + usv, this allows executing across multiple chains in devnet env (namechain)


adraffy (approved):
Does this need a flag to not deploy it in the case it doesn't exist?

TateB:
> Does this need a flag to not deploy it in the case it doesn't exist?

yeah i suppose i can remove the flag check entirely, and just check the deployed bytecode

deploy/utils/shared/00_deploy_multicall.ts
```diff
@@ -0,0 +1,93 @@
...
+        bytecode: '0x',
+        metadata: '',
+      },
+      expectedAddress: multicallAddress,
+    })
+
+    const multicallDeployedBytecode = await viem.publicClient.getCode({
+      address: multicallAddress,
+    })
+    if (!multicallDeployedBytecode) throw new Error('Multicall3 not deployed')
```
adraffy, line 85: This will never fail because it returns `0x`?

adraffy (commented):
I was saying it should have a flag to not deploy it, but I guess that can be done by just not using the script.

deploy/utils/shared/00_deploy_multicall.ts
TateB: viem's `getCode` returns undefined if `0x`

PR #498, opened 2025-10-22
Improve Reverse Resolver Introspection

adraffy:
* bumped `@unruggable/gateways` `@namestone/ezccip` and `ethers`
* expanded [INameReverser](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/contracts/reverseResolver/INameReverser.sol) &mdash; now sufficient to identify a "reverse" resolver
* added `IVerifiableResolver` to [ChainReverseResolver](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/contracts/reverseResolver/ChainReverseResolver.sol#L70-L78)
* reduced some duplicated constants
* added [fuzz](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/test/reverseResolver/TestChainReverseResolver.ts#L260-L288) test


contracts/reverseResolver/ChainReverseResolver.sol
```diff
@@ -1,21 +1,27 @@
...
 import {INameReverser} from "./INameReverser.sol";
+import {ENSIP19} from "../utils/ENSIP19.sol";
 
 /// @title Chain Reverse Resolver
 /// @notice Reverses an EVM address using the first non-null response from the following sources:
-///         1. `L2ReverseRegistrar` on L2 chain via Unruggable Gateway
-///         2. `IStandaloneReverseRegistrar` for "default.reverse"
+///
+/// 1. `L2ReverseRegistrar` on L2 chain via Unruggable Gateway
+/// 2. `IStandaloneReverseRegistrar` for "default.reverse"
```
TateB, line 20: why the change?

TateB (approved):

contracts/reverseResolver/ChainReverseResolver.sol
adraffy: The indent removal?  I think the indent is weird when the docstring spans more than a couple lines.

However, I think all of the docstrings should be standardized closer to release, regarding `@dev` vs `@notice` use, formatting, etc.

TateB: sure we can change to remove multiline indentation, i'll try to write a full style guide sometime this week. does mean we need to revert the indentation elsewhere though.

on `@dev` and `@notice`, the distinction should be:
- `@notice`: documentation for a user of a contract, after deployment, e.g. interacting with deployed contract
- `@dev`: documentation for a user (or developer) of a contract, before deployment, e.g. working inside the codebase, utilising a library, using contract inheritance

adraffy: I think for a lot of situations, that line is hard to draw.   I feel like I'm just randomly picking one or the other in many situations.  Increasingly, I want to use `@notice` for everything, and only use `@dev` for private functions or nerdy things like precomputed `selectors` and `interfaceId`'s.

I looked through other codebases and found pretty arbitrary use of these tags.

* This one [seems okay](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L381-L384) as the `@dev` tells you about the internal details.

* This function is [effectively private](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L552-L553) so the comment is `@dev`.  I didn't include `@param` or `@return` because it's just noise, as the comment already explains exactly what it's doing.

* This one [looks wrong](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L306) as I think that should be `@notice` at the top, and maybe the `@param`  moved below the large comment block.

* This one [looks wrong too](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L213-L220) as the examples should probably be above, in `@notice`.

* This [event](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L113-L115) follows my thinking above.  Simiarlly, these [constants](https://github.com/ensdomains/namechain/blob/feat/eth-tld-resolver-wrapper/contracts/src/L1/resolver/ETHTLDResolver.sol#L80-L84) are private and technical.

TateB: how about if `@notice` is for describing the functions external use, and `@dev` is for the internal workings?

Addausi:
> * bumped `@unruggable/gateways` `@namestone/ezccip` and `ethers`
> * expanded [INameReverser](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/contracts/reverseResolver/INameReverser.sol) — now sufficient to identify a "reverse" resolver
> * added `IVerifiableResolver` to [ChainReverseResolver](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/contracts/reverseResolver/ChainReverseResolver.sol#L70-L78)
> * reduced some duplicated constants
> * added [fuzz](https://github.com/ensdomains/ens-contracts/blob/feat/bet-497-urg/test/reverseResolver/TestChainReverseResolver.ts#L260-L288) test

PR #503, opened 2025-11-25
ENSIP-24 Resolver Profile for Arbitrary Data Resolution

clowestab:
- Added ENSIP-24 Interfaces x2
- Added basic mixin
- Added basic test coverage
- Updated README


adraffy (changes requested):
Create a separate mixin for the iterable `ISupportedDataKeys`.  

`PublicResolver` should only implement the non-iterable mixin (just `IDataResolver`).

Both mixins should support versioning.

Once merged, I'll add support in ENSv2 ~~`DedicatedResolver`~~ [OwnedResolver](https://github.com/ensdomains/contracts-v2/blob/main/contracts/src/resolver/OwnedResolver.sol)

clowestab:
Separated mixins as requested. Added an `_afterSetData()` hook to `DataResolver` to connect the two.
Added versioning.

adraffy (approved):

contracts/resolvers/profiles/IDataResolver.sol
```diff
@@ -0,0 +1,22 @@
...
+pragma solidity >=0.8.4;
+
+/// @dev Interface selector: `0xecbfada3`
+interface IDataResolver {
+    /// @notice For a specific `node`, the data associated with a `key` has changed.
+    event DataChanged(
+        bytes32 indexed node, 
+        string indexed indexedKey,
+        string key, 
+        bytes indexed indexedData
```
makoto, line 11: Does this need to be indexed? If you index bytes, you can't decode the full bytes (only hashed version in the topic)

contracts/resolvers/profiles/ISupportedDataKeys.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.4;
+
+/// @dev Interface selector: `0x29fb1892`
+interface ISupportedDataKeys {
+    /// @notice For a specific `node`, get an array of supported data keys.
+    /// @param node The node (namehash).
+    /// @return The keys for which we have associated data.
+    function supportedDataKeys(bytes32 node) external view returns (string[] memory);
+}
```
makoto: Missing newline at end of file

contracts/resolvers/profiles/SupportedDataKeys.sol
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    function supportsInterface(
+        bytes4 interfaceID
+    ) public view virtual override returns (bool) {
+        return
+            interfaceID == type(ISupportedDataKeys).interfaceId ||
+            super.supportsInterface(interfaceID);
+    }
+}
```
makoto: Missing newline at end of file

contracts/resolvers/profiles/IDataResolver.sol
```diff
@@ -0,0 +1,22 @@
...
+    
+    /// @notice For a specific `node`, get the data associated with the key, `key`.
+    /// @param node The node (namehash) for which data is being fetched.
+    /// @param key The key.
+    /// @return The associated arbitrary `bytes` data.
+    function data(
+        bytes32 node,
+        string calldata key
+    ) external view returns (bytes memory);
+}
```
makoto: Missing newline at end of file

contracts/resolvers/profiles/IDataResolver.sol
makoto: I also wondered if this is for gas saving, but if so then probably indexedData is not worth adding in the event as you can query onchain. I don't know the case when you want to query indexedData by its hash

makoto (changes requested):
Can you get rid of bun.lockb as we have bun.lock?

contracts/resolvers/profiles/IDataResolver.sol
adraffy: I think there was some discussion on this and it matches whats in https://docs.ens.domains/ensip/24/ 

It lets you search for the hash of a known value.

adraffy:
Ok wait, not fixed, I can't push into that repo.

PR #509, opened 2025-12-25
feat: Replace EllipticCurve with EIP-7951 P-256 precompile for Algorithm 13

estmcmxci:
- Add P256Precompile library following ModexpPrecompile pattern
- Update P256SHA256Algorithm to use native precompile (0x100)
- Remove EllipticCurve dependency and BytesUtils usage
- Use inline assembly for calldata slicing instead of BytesUtils (more gas-efficient, no memory allocations needed)
- Gas reduction: ~200k+ gas → ~3,500 gas per verification (~98% savings)

Note: TestAlgorithms.test.ts will fail on local networks without EIP-7951 activation. Tests pass on networks with precompile (Sepolia, Mainnet after Fusaka upgrade). Working test suite with mocked precompile available at: https://github.com/eurekaetcetera/dnssec-solutions/tree/master/ens-p256-test

PR #513, opened 2026-01-19
Update hardhat to 3.1.4

makoto:
Upgrade the hardhat to the latest which includes fusaka upgrade (pre requisite to merge https://github.com/ensdomains/ens-contracts/pull/509)


adraffy (approved):
I don't think there's any issue here but tate would know more.

PR #514, opened 2026-01-20
Consider adding slippy to lint Solidity files

fvictorio:
Hi, I've been working for a while on [Slippy](https://slippy-lint.github.io/slippy/), a new linter for Solidity. This PR adds it to the ENS repo, enables a few of its rules, and fixes what they report.

Since the fixes are useful on their own, I split the commits so that Slippy is actually added in the last one. That way, if you are interested in the fixes but not in adding the linter, you can drop that last commit and merge the rest.


fvictorio:
Closing this for now to avoid polluting the list of open PRs, happy to re-open if there's interest in this.

PR #517, opened 2026-01-30
Add RegistrarSecurityController and RootSecurityController for security council access

Arachnid:
These contracts facilitate the security council immediately disabling a vulnerable registrar controller or TLD in case of compromise, so the DAO can act to fix and reinstate the vulnerable code later.

### DIFF
--- a/contracts/ethregistrar/RegistrarSecurityController.sol
+++ b/contracts/ethregistrar/RegistrarSecurityController.sol
@@ -1,123 +1,52 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-import "./BaseRegistrarImplementation.sol";
-import "@openzeppelin/contracts/access/Ownable.sol";
+import {BaseRegistrarImplementation} from "./BaseRegistrarImplementation.sol";
+import {Controllable} from "../root/Controllable.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 
-/**
- * @title RegistrarSecurityController
- * @notice Break-glass controller for the base registrar.
- * @dev Acts as a pass-through for the base registrar, but with the ability for
- *      security controllers to disable registrar controllers.
- */
-contract RegistrarSecurityController is Ownable {
-    bytes4 private constant INTERFACE_META_ID =
-        bytes4(keccak256("supportsInterface(bytes4)"));
+/// @title RegistrarSecurityController
+/// @notice Break-glass controller for the base registrar.
+/// @dev Acts as a pass-through for the base registrar, but with the ability for
+///      security controllers to disable registrar controllers.
+contract RegistrarSecurityController is Controllable, ERC165 {
 
-    /**
-     * @notice Emitted when a security controller is added.
-     * @param controller The security controller address.
-     */
-    event ControllerAdded(address indexed controller);
-    /**
-     * @notice Emitted when a security controller is removed.
-     * @param controller The security controller address.
-     */
-    event ControllerRemoved(address indexed controller);
-
-    /**
-     * @notice Security controllers authorized to disable registrar controllers.
-     * @dev These addresses can call `disableRegistrarController`.
-     */
-    mapping(address => bool) public controllers;
-
-    /**
-     * @notice The registrar this controller manages.
-     */
+    /// @notice The registrar this controller manages.
     BaseRegistrarImplementation public registrar;
 
-    /**
-     * @dev Restricts actions to security controllers.
-     */
-    modifier onlyController() {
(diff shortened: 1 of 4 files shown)


contracts/ethregistrar/RegistrarSecurityController.sol
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "./BaseRegistrarImplementation.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
```
TateB: nit: use triple slash natspec

contracts/ethregistrar/RegistrarSecurityController.sol
```diff
@@ -0,0 +1,123 @@
...
+ * @title RegistrarSecurityController
+ * @notice Break-glass controller for the base registrar.
+ * @dev Acts as a pass-through for the base registrar, but with the ability for
+ *      security controllers to disable registrar controllers.
+ */
+contract RegistrarSecurityController is Ownable {
+    bytes4 private constant INTERFACE_META_ID =
+        bytes4(keccak256("supportsInterface(bytes4)"));
```
TateB: nit: import and inherit erc165 instead of defining manually

contracts/ethregistrar/RegistrarSecurityController.sol
TateB, line 1: we already have `Controllable` implementations that uses a `ControllerChanged` event. those implementations do use a string revert instead of a custom error, but in this implementation (in `onlyController`) there is no error at all.

i'd say it's worth:
- using the already defined `ControllerChanged` event pattern
- adding a custom error such as `ControllerUnauthorizedAccount(address account)` for the modifier

contracts/ethregistrar/RegistrarSecurityController.sol
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "./BaseRegistrarImplementation.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
```
TateB: nit: use v5 ownable from `@openzeppelin/contracts-v5`

contracts/ethregistrar/RegistrarSecurityController.sol
TateB: looks like the controller code is mostly from the base registrar itself, but since we're using the other event pattern more often i'd still say it's worth doing

contracts/ethregistrar/RegistrarSecurityController.sol
```diff
@@ -0,0 +1,123 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import "./BaseRegistrarImplementation.sol";
```
TateB: nit: use explicit imports

test/ethregistrar/TestRegistrarSecurityController.test.ts
```diff
@@ -0,0 +1,253 @@
...
+          {
+            account: accounts[1],
+          },
+        ),
+      ).toBeRevertedWithString('Ownable: caller is not the owner')
+    })
+  })
+
+  describe('supportsInterface', () => {
+    it('should support ERC165', async () => {
```
TateB: interfaces can be tested at the top of the tests:
```ts
  shouldSupportInterfaces({
    contract: () =>
      loadFixture().then(({ registrarSecurityController }) => registrarSecurityController),
    interfaces: ['IERC165'],
  })
```

TateB (changes requested):

adraffy (commented):
I had similar questions about `INTERFACE_META_ID` and using the imported interfaces.

My understanding is:
* registry.owner("eth") = NFT
* NFT.owner() = this
* this has owner
* owner can set registry.resolver("eth")
* owner can set new NFT.owner()
* owner can add/remove NFT controllers
* owner can add/remove controllers
* controllers can remove NFT controllers

Which seems straightforward in the code.

contracts/ethregistrar/RegistrarSecurityController.sol
Arachnid: Right, I copied it over for consistency with the base registrar. Happy to change it though.

contracts/ethregistrar/RegistrarSecurityController.sol
Arachnid: I can't, as `context` from the legacy contracts is imported elsewhere, and the two versions conflict.

Arachnid:
@TateB PTAL

TateB (approved):

PR #537, opened 2026-03-18
chore: sepolia dev deployment + update rocketh

TateB:
changes:
- `sepolia-dev` deployment, intended for use with v2.
- updated rocketh and hardhat deploy
- modified deploy scripts such that artifacts/abis are referenced directly by their generated file. this massively improves typescript performance across the repo: in-ide, typechecks, `hardhat deploy` calls, etc.
- tweaked psl/set tlds deploys to broadcast more txs per block (only on testnet fresh deploys), speeds up deploy


tsconfig.json
```diff
@@ -12,7 +12,9 @@
     "resolveJsonModule": true,
     "types": ["vitest/globals"],
     "paths": {
```
jore0-cell, line 14:
```suggestion
    "paths": {
```

PR #552, opened 2026-06-14
refactoring

### LINKED ISSUE #8: Accept stablecoins on the .eth registrar controller
One option for this would be to integrate a router such as the uniswap router, so users can pay in *any* currency.

brenzi:
This breaks with some ENS legacy:

* fully remove NameWrapping (EIP-1155) 
* subnames get their own non-token registrar
* do not rely on and subgraph indexer or NFT rendering service. rendering, enumerate and lookup fully onchain
* limit 2LD length. closes #8


brenzi:
wrong target :-(

